PR #7, opened 2018-05-02
Add cli parsing

### LINKED ISSUE #2: port code to use clap and clap-derive or docopt
Right now we don't support command-line arguments:

https://github.com/rust-lang-nursery/borrow-check/blob/8ef88dbe968db92dd9dfa0c0296c3c4a57608e0d/src/cli.rs#L9-L11

I'd like to add clap (with clap-derive) or docopt. Then we can have some command-line options. 

cc @killercup -- what's the "new hotness" in terms of super simple CLI wrappers? =)

qmx:
this attempts to fix #2 - introduces [structopt](https://github.com/TeXitoi/structopt), and provides the options suggested by @nikomatsakis 

I didn't explore adding exclusion groups in lieu of not trying to read beyond what was specified in the issue :)


nikomatsakis:
Just realized we don't have travis setup =)

PR #8, opened 2018-05-02
add Algorithm enum

### LINKED ISSUE #3: add multiple "engines"
Right now, when you invoke `Output::compute`, we *always* use the naive differential-dataflow implementation to do that computation:

https://github.com/rust-lang-nursery/borrow-check/blob/8ef88dbe968db92dd9dfa0c0296c3c4a57608e0d/src/output/mod.rs#L33-L35

I would like to pass in an enum and select from multiple options. In particular, the current should be retained, with the title "naive". Then I imagine we can add more optimized variants. This would probably build on #2 to allow for a command-line option that selects which algorithm to use.

qmx:
this attempts to fix #3 

once #7 lands I'll rebase this one


nikomatsakis:
@qmx needs rebase =)

qmx:
@nikomatsakis done

src/cli.rs
```diff
@@ -25,9 +25,9 @@ pub fn main(opt: Opt) -> Result<(), Error> {
             let tables = &mut intern::InternerTables::new();
 
             let result: Result<(Duration, Output), Error> = do catch {
-                let all_facts = tab_delim::load_tab_delimited_facts(tables, &facts_dir)?;
                 let verbose = opt.verbose;
-                timed(|| Output::compute(all_facts, verbose))
+                let all_facts = tab_delim::load_tab_delimited_facts(tables, &facts_dir)?;
+                timed(|| Output::compute(all_facts, Algorithm::Naive, verbose))
```
nikomatsakis: can we add this algorithm choice to the `Opt` structure somehow?

PR #14, opened 2018-05-02
Add support for outputting to files

bobtwinkles:
Does pretty much exactly what it says on the tin :smile:. Adds support for dumping the results to output files. We still dump to stdout by default, but now if `-o` or `--output` is provided that directory is used as the output directory. Different sets of facts are dumped to different files.

### DIFF
--- a/src/output/mod.rs
+++ b/src/output/mod.rs
@@ -14,6 +14,7 @@ use crate::intern::InternerTables;
 use fxhash::FxHashMap;
 use std::borrow::Cow;
 use std::collections::{BTreeMap, BTreeSet};
+use std::io::{self, Write};
 
 mod dump;
 mod timely;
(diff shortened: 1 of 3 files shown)


src/output/mod.rs
```diff
@@ -48,13 +50,37 @@ impl Output {
...
+        return Ok(());
+
+        fn writer_for(out_dir: &Option<PathBuf>, name: &str) -> io::Result<Box<Write>> {
+            // create a writer for the provided output.
+            // If we have an output directory use that, otherwise just dump to stdout
+            use std::fs;
+            use std::path::Path;
+
+            Ok(match out_dir {
+                Some(ref dir) => {
```
nikomatsakis: Nit: you don't need the `ref` here, as you are matching a `&` value

src/output/mod.rs
```diff
@@ -48,13 +50,37 @@ impl Output {
...
+                    if !dir.exists() {
```
nikomatsakis: Nit: you can just use `fs::create_dir_all(&dir)?`, I think

PR #16, opened 2018-05-04
Add tracking for in/out degree

### LINKED ISSUE #15: gather some statistics
There are a few things I'd love to know about the clap-rs input:

For example, what are the max in- and out-degrees that we observe in the subset graph at any point?

Similarly, when a region does dead, what are the in- and out-degrees it has at that time?

bobtwinkles:
Fixes #15. The multidegree tracking isn't critical and I can rip it out pretty easily, but I wanted  to answer @chrisvittal's [question on gitter](https://gitter.im/rust-impl-period/WG-compiler-nll?at=5aebbb5603cafa797b25eb98) :wink: (the answer turns out to be yes).

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -53,6 +53,17 @@ pub fn main(opt: Opt) -> Result<(), Error> {
                         let seconds: f64 = duration.as_secs() as f64;
                         let millis: f64 = duration.subsec_nanos() as f64 * 0.000_000_001_f64;
                         println!("Time: {:0.3}s", seconds + millis);
+
+                        if opt.verbose {
+                            println!("Max region graph in/out-degree: {} {}",
+                                     output.region_degrees.max_in_degree(),
+                                     output.region_degrees.max_out_degree());
+                            if output.region_degrees.has_multidegree() {
+                                println!("Found multidegree");
+                            } else {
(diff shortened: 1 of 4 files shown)


bobtwinkles:
Ooops, looks like Travis is sad because of rust-lang/rust#50401. I'll submit a PR to fix that.

nikomatsakis:
@bobtwinkles this seems good -- I was thinking last night that this could be done in differential-dataflow itself pretty nicely, instead of writing up custom rust code. If you're curious to learn how that stuff works, I can give you a few hints. The idea would be roughly to add some code like:

```rust
let region_in_degrees = subset
    .map(|(_r1, r2, p)| (r2, p)) // just preserve target region and point
```

this will create numerous duplicate tuples; the differential-dataflow preserves that and tracks their number. So then when you invoke `inspect_batch`, you will get a "multiplicity" that is equal to the in-degree. (If you look at the existing `inspect_batch` routines, you will see we always assert the multiplicity is 1; this is because we invoke `distinct()` to remove multiple instances.)

There's probably a way to find the count or the max as well, I'm not sure off the top of my head. Anyway, as this is just debug output, it doesn't matter that much, but if you want to take a stab at refactoring, might be fun for getting your feet wet.

PR #18, opened 2018-05-04
adopt extern-prelude and extern-absolute-path features

nikomatsakis:
Now we don't need the `extern crate`. Though the need to pull in "private" macros from clap is a bit lame, as is suddent appearance of "deprecated method" warnings.


nikomatsakis:
Oh, actually, that warning just seems to be occurring from the newer nightly.

Issue #20, opened 2018-05-10
Idea: condense CFG


nikomatsakis:
Currently we have an awful lot of points. However, it is rare that we need so many. I'm not sure of the *exact* conditions when we can compress two points, but it seems clear that we can. For example, if we have two points P and Q where 

- P has one successor (Q)
- Q has one predecessor (P)
- The set of live regions at P is equal to set of live regions at Q
- There are no outlives relations at P or Q

then I think pretty clearly the subset relation would be the same during both of them.

We often know the first condition (one succ, one pred) just structurally: only the *start* of a basic block can have multiple preds, and only the *mid* point of a terminator can have multiple successors. The set of live regions is also easily approximated using the set of live *variables*, I imagine.

PR #23, opened 2018-05-11
Detransitize the differential dataflow computation

nikomatsakis:
This introduces a new variant of the differential dataflow computation. This variant attempts to avoid computing the transitive closure of subset relations -- instead of computing the full TC at each point, it rather computes the TC along edges, and only for the dead regions along each edge. We also switch to using `distinct_total` instead of `distinct`, which is faster. The naive code is retained and used for comparison. The new method seems to produce the same results as the old.

Current results for clap on my machine:

| Version | Timing |
| --- | --- |
| Before | 131s |
| Without `distinct_total` | 63s |
| With `distinct_total` | 41.239s |

Progress!

### DIFF
--- a/src/output/timely_opt.rs
+++ b/src/output/timely_opt.rs
@@ -10,9 +10,10 @@
 
 //! Timely dataflow runs on its own thread.
 
-use crate::facts::AllFacts;
+use crate::facts::{AllFacts, Point};
 use crate::output::Output;
 use differential_dataflow::collection::Collection;
+use differential_dataflow::operators::iterate::Variable;
 use differential_dataflow::operators::*;
 use std::collections::{BTreeMap, BTreeSet};
 use std::mem;
@@ -21,8 +22,23 @@ use std::sync::Arc;
 use std::sync::Mutex;
 use timely;
 use timely::dataflow::operators::*;
+use timely::dataflow::Scope;
 
-pub(super) fn compute(dump_enabled: bool, all_facts: AllFacts) -> Output {
+pub(super) fn compute(dump_enabled: bool, mut all_facts: AllFacts) -> Output {
+    // Declare that each universal region is live at every point.
+    let all_points: BTreeSet<Point> = all_facts
+        .cfg_edge
+        .iter()
+        .map(|&(p, _)| p)
+        .chain(all_facts.cfg_edge.iter().map(|&(_, q)| q))
+        .collect();
+
+    for &r in &all_facts.universal_region {
+        for &p in &all_points {
+            all_facts.region_live_at.push((r, p));
+        }
+    }
+
     let result = Arc::new(Mutex::new(Output::new(dump_enabled)));
 
     // Use a channel to send `all_facts` to one worker (and only one)
@@ -57,7 +73,6 @@ pub(super) fn compute(dump_enabled: bool, all_facts: AllFacts) -> Output {
                 let_collections! {
                     let (
                         borrow_region,
-                        universal_region,
                         cfg_edge,
                         killed,
                         outlives,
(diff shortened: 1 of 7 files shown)


nikomatsakis:
I checked manually and the new analysis also produces the same results on the clap example as the old one.

src/output/timely_opt.rs
```diff
@@ -0,0 +1,330 @@
...
+                                .map(|((_r2, p), (r1, q), r3)| (r1, r3, p, q))
+                        };
+
+                        dead_can_reach.set(&dead_can_reach_base
+                            .concat(&dead_can_reach2)
+                            .distinct_total())
+                    };
+
+                    // subset(R1, R2, Q) :-
+                    //   subset(R1, R2, P) :-
```
lqd: tiny typo: should be "," instead of ":-"

src/output/timely_opt.rs
```diff
@@ -0,0 +1,330 @@
...
+                            }
+                        }
+                    });
+
+                    subset.inspect_batch({
+                        let result = result.clone();
+                        move |_timestamp, facts| {
+                            let mut result = result.lock().unwrap();
+                            for ((r1, r2, location), _timestamp, multiplicity) in facts {
+                                assert_eq!(*multiplicity, 1);
```
lqd: BTW this assert triggers with `-v``--stats` — apparently `multiplicity` being -1 here — with both input data sets.

frankmcsherry: I just looked into this, and it seems that the problem goes away if you put a `consolidate()` after the definition of `subset`. This coalesces all updates together, and .. it seems to mean that the final results of `subset` are non-negative.

I think what happened here (testing now, but 5 min compile loop) is that 
```rust
                    let subset = Variable::from(subset_base.clone());
``` 
provides an initial value for `subset` and then removes it in the next round. The code concats the collection back in, so these two changes logically cancel, but the removal does produce negative tuples before they get cancelled.

I'm trying out 
```rust
                    let subset = Variable::from(subset_base.clone().filter(|_| false));
```
now which just says "don't start with any facts", which will avoid removals and shouldn't change the results as `subset_base` is concatenated in to the recursive definition of `subset` anyhow.

frankmcsherry: Yes, adding in the filter to prevent initial values going in seems to prevent the assert, even after removing the `consolidate()` that I added. Now, technically speaking I haven't seen it assert so I'm trying that now to make sure that the code is actually being called (though using `--stats` and seeing stats outputs, so probably?).

frankmcsherry: Yup, without the filter I do see the assert, and either the filter or consolidate seem to prevent it.

nikomatsakis:
I haven't had time to read the comments yet, but I pushed some new commits that also remove the transitive loan propagation. This brings total time down to 15s (without altering final output).

nikomatsakis:
I pushed a few minor commits (e.g., using `join_map` instead of `join` and `map` etc), but they didn't affect overall runtime (not too surprising). Still, I believe @frankmcsherry that using `join_map` is more efficient in principle? (Perhaps once we make more strides =)

At this point I *think* the best thing to do is to pursue #20 (which is blocked on https://github.com/rust-lang-nursery/borrow-check/issues/24)

src/output/timely_opt.rs
nikomatsakis: Would you recommend I do this `filter` trick then? Or is `consolidate` better?

nikomatsakis: (Could I just create an empty collection some other way...?)

nikomatsakis: I'm sort of confused why we get the negative multiplicities, but I guess it's sort of an "internal accounting" thing?

frankmcsherry: I think `filter` is probably "better". I forgot that with the `filter` you can drop the `clone`. It only really makes a difference if the input collection is large (the difference is one additional positive and negative instance of each input fact).

This is all here because `Variable` supports non-monotonic variables, and in principle a strictly monotonic variable shouldn't need any `from()` input.

frankmcsherry: The negative multiplicities are really just about being general enough for non-monotonic computation. If you want to compute f^i(x) for a general f, we need to start with x, compute f(x), but then subtract x to get f(x) - x. That change is what needs to flow back around the loop to keep the incremental computation going correctly. 

In the case of Datalog you wouldn't need to subtract off the x, because you know you'll have distincts everywhere, but more general computation needs the -x to correctly determine f^i(x) rather than f(x + f(x + f(x + ...))).

frankmcsherry: Also, yes I should totally have a better way to create an empty collection. Filter should be pretty cheap (the closure should optimize down to no per-record work). Probably the best other way is to use the same `to_stream` stuff you are using for inputs, but just with a `None` as the input `IntoIterator`.

frankmcsherry:
> Still, I believe @frankmcsherry that using join_map is more efficient in principle? (Perhaps once we make more strides =)

Yup, it will save a bunch of copies, mostly. At some point I'll show you how to use `join_core` which is what lies underneath `join`, `join_map`, and `semijoin`, and which allows you to put arbitrary iterator logic in the output production. In principle you can fuse all of the maps into their preceding joins, except `antijoin` which has that concat there.

src/output/timely_opt.rs
nikomatsakis: Yeah I figured I would do `None` or `vec![]` with an as-stream or whatever.

nikomatsakis:
OK, I pushed the change to start with empty variables and grow them. To be honest, I find the code cleaner this way, and I manually verified that the asserts don't fire when using `-v`.

nikomatsakis:
It also seems to either make things mildly faster =)

Issue #4, opened 2018-05-01
add `Invalidated` tuples


nikomatsakis:
This requires modifying both rustc and borrow-check. The idea is to add the `Invalidated` tuples described [in my blog post](http://smallcultfollowing.com/babysteps/blog/2018/04/27/an-alias-based-formulation-of-the-borrow-checker/#defining-an-error). We would begin by modifying rustc [as described in this gitter log](https://gitter.im/rust-impl-period/WG-compiler-nll?at=5ae8c9e693dc5e5cdb43cc22):

- Today, after computing the borrows in scope, we walk the MIR and invoke `check_access_for_conflict`:
  - that walks over the loans in scope and compares the accessed path against the borrowed path
- Instead, we would walk the MIR ahead of time, before "borrows in scope" are computed:
  - for each access, we would walk **all** loans and do the same comparison
  - if the access would conflict with the loan, we generate a `invalidated(L, P)` fact

This will require some refactoring of borrow check to extract the relevant code. Eventually, I think this can be made extra efficient as well, but basically indexing the paths borrowed by a loan. But we can begin with simple stuff and worry about that later.

Then we would have to extend the `AllFacts` definitions in rustc and this repo to include `Invalidated` facts.

Once those facts are available, we should open a separate issue to discuss how best to make use of them.

nikomatsakis:
cc @bobtwinkles @KiChjang -- I thought one of you two might be up for this. I think that adding this support to rustc is pretty high priority because it's likely a good "enabler" for further improvements.

nikomatsakis:
also possibly @spastorino

spastorino:
I can tackle this

nikomatsakis:
So @bobtwinkles made good progress the other day in [this rustc branch](https://github.com/bobtwinkles/rust/tree/nll_facts_invalidate). I'm now looking a bit past that point --

It seems like we would want to do a few things here, and we can play around with them now by synthesizing the "Facts" files.

**Step 1.** First, we need to add a `invalidated` field to `AllFacts`:

https://github.com/rust-lang-nursery/borrow-check/blob/4945c130cb1f3af7c072dc9856b6fe21b3a9f1f7/src/facts.rs#L3-L5

Something like

```rust
invalidated: Vec<(Loan, Point)>
```

then we would need to make that compile loading facts from `invalidated.facts`. 

[...]

PR #28, opened 2018-05-14
Add invalidates facts

spastorino:
I'll leave this here as a discussion starting point :).

Related to #4

### DIFF
--- a/src/facts.rs
+++ b/src/facts.rs
@@ -22,6 +22,9 @@ crate struct AllFacts {
 
     /// `region_live_at(R, P)` when the region R appears in a live variable at P
     crate region_live_at: Vec<(Region, Point)>,
+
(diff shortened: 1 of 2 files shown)


nikomatsakis:
Can you add that to the list of things to add here

https://github.com/rust-lang-nursery/borrow-check/blob/4945c130cb1f3af7c072dc9856b6fe21b3a9f1f7/src/tab_delim.rs#L33-L42

probably we will need to `touch` the various files as a result

spastorino:
@nikomatsakis unsure what's going on, but I have added that and see when I click on commit. Unsure why you don't see the change.

nikomatsakis:
oh, I must be blind. I see it now =)

spastorino:
This can be closed, is superseded by https://github.com/rust-lang-nursery/polonius/pull/34

Issue #25, opened 2018-05-14
allow command line arguments to be given to differential dataflow


nikomatsakis:
Right now, when we invoke differential-dataflow, we give in no command line arguments:

https://github.com/rust-lang-nursery/borrow-check/blob/4945c130cb1f3af7c072dc9856b6fe21b3a9f1f7/src/output/timely_opt.rs#L51

As you can see in [the docs.rs page](https://docs.rs/timely/0.5.1/timely/execute/fn.execute_from_args.html), though, there are a number of arguments for things like using multiple threads. I actually think it's better if we avoid threads by default, but it'd be interesting to see what effects those flags have. We should extend our CLI to permit us to pass in "timely dataflow" flags, and then thread them through to that line above.

We can also pass them to the naive algorithm, here:

https://github.com/rust-lang-nursery/borrow-check/blob/4945c130cb1f3af7c072dc9856b6fe21b3a9f1f7/src/output/naive.rs#L34

nikomatsakis:
Maybe we should just extend our CLI with the `-w` switch to indicate how many worker threads to use.  I don't think we want to mess with multiple processes.

lqd:
Note: most of #30 was removed in the switch-to-datafrog PR, and maybe we can close this issue ?

PR #30, opened 2018-05-15
added support to pass argument to  differential dataflow

PramodBisht:
related to #25 
For a now, this is implemented only for worker argument of timely.
@nikomatsakis  please check.

### DIFF
--- a/src/output/naive.rs
+++ b/src/output/naive.rs
@@ -22,7 +22,7 @@ use std::sync::Mutex;
 use timely;
 use timely::dataflow::operators::*;
 
-pub(super) fn compute(dump_enabled: bool, all_facts: AllFacts) -> Output {
+pub(super) fn compute(dump_enabled: bool, all_facts: AllFacts, workers: u32) -> Output {
     let result = Arc::new(Mutex::new(Output::new(dump_enabled)));
 
     // Use a channel to send `all_facts` to one worker (and only one)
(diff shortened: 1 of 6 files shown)


src/output/naive.rs
```diff
@@ -22,16 +22,21 @@ use std::sync::Mutex;
...
 
     // Use a channel to send `all_facts` to one worker (and only one)
     let (tx, rx) = mpsc::channel();
     tx.send(all_facts).unwrap();
     mem::drop(tx);
     let rx = Mutex::new(rx);
+    let mut dataflow_arg = Vec::new();
+    if workers > 1 {
+        let item = format!("{}{}", "w", workers);
+        dataflow_arg.push(item);
```
nikomatsakis: I think this should be

```
dataflow_arg.push(format!("-w"));
dataflow_arg.push(format!("{}", workers));
```

though one thing may work too

src/output/timely_opt.rs
```diff
@@ -47,8 +47,13 @@ pub(super) fn compute(dump_enabled: bool, mut all_facts: AllFacts) -> Output {
     tx.send(all_facts).unwrap();
     mem::drop(tx);
     let rx = Mutex::new(rx);
+    let mut dataflow_arg = Vec::new();
+    if workers > 1 {
+        let item = format!("{}{}", "w", workers);
+        dataflow_arg.push(item);
```
nikomatsakis: same here; maybe we want a helper fn :)

PramodBisht: for helper function, any ideal place where I should store this? or should I create common helper functions file for that?

src/output/naive.rs
PramodBisht: done!

src/output/timely_opt.rs
PramodBisht: @nikomatsakis done!

PR #32, opened 2018-05-16
Add location insensitive analysis

### LINKED ISSUE #29: location-insensitive first pass
Building on my notes from #4 (cc @spastorino @bobtwinkles), as well as the promising results from https://github.com/rust-lang/rust/pull/50593, I was thinking that one way to get things faster would be to begin with a "coarser-grained" analysis and then -- only for those loans that resulted in an error -- do the full analysis.

This will build on #4 -- we need the set of `Invalidated` facts to make this work, but it's a separate body of work. That is, we can implement this 'location-insensitive' analysis as an experimental pre-pass and start to measure it etc.

Roughly speaking, the idea would be to first create a "location-insensitive" subset relation. Unlike the current rules, we would never "remove" things from this relation. The rules would be very simple, I think actually just this:

```
subset(R1, R2) :- outlives(R1, R2, _P).
```

Then we would propagate loans along this subset relation, again ignoring location:

```
requires(R1, L1) :- borrow_region(R1, L1, _P).
requires(R2, L1) :- requires(R1, L1), subset(R1, R2).
```

In fact, even these very simple rules are a kind of transitive closure, and hence we might find that we want to speed them up. But let's assume it's fast enough for now.

Now we can figure out which loans **might** be invalidated at each point:

```
potential_errors(L, P) :-
  invalidated(L, P),
  requires(R, L),
  region_live_at(R, P).
```

Once we have the set of *potential errors*, we can use those to target the more precise analysis — but now we know that we only care about the loans where we found *potential errors*.

spastorino:
Part of #29 

Performance difference between this algorithm, the TimelyOpt and the Naive one for the following clap-rs example ...

```
[santiago@archlinux borrow-check (location_insensitive)]$ cargo run --release -- --skip-tuples inputs/clap-rs/app-parser-\{\{impl\}\}-add_defaults/ | head
    Finished release [optimized] target(s) in 0.07s
     Running `target/release/borrow-check --skip-tuples 'inputs/clap-rs/app-parser-{{impl}}-add_defaults/'`
--------------------------------------------------
Directory: inputs/clap-rs/app-parser-{{impl}}-add_defaults/
Time: 117.131s
[santiago@archlinux borrow-check (location_insensitive)]$ cargo run --release -- --skip-tuples -a TimelyOpt inputs/clap-rs/app-parser-\{\{impl\}\}-add_defaults/ | head
    Finished release [optimized] target(s) in 0.10s
     Running `target/release/borrow-check --skip-tuples -a TimelyOpt 'inputs/clap-rs/app-parser-{{impl}}-add_defaults/'`
--------------------------------------------------
Directory: inputs/clap-rs/app-parser-{{impl}}-add_defaults/
Time: 16.729s
[santiago@archlinux borrow-check (location_insensitive)]$ cargo run --release -- --skip-tuples -a LocationInsensitive inputs/clap-rs/app-parser-\{\{impl\}\}-add_defaults/ | head
    Finished release [optimized] target(s) in 0.07s
     Running `target/release/borrow-check --skip-tuples -a LocationInsensitive 'inputs/clap-rs/app-parser-{{impl}}-add_defaults/'`
--------------------------------------------------
Directory: inputs/clap-rs/app-parser-{{impl}}-add_defaults/
Time: 2.058s
```

~57x faster than Naive for this example.
~8x faster than TimelyOpt for this example.

### DIFF
--- a/src/output/location_insensitive.rs
+++ b/src/output/location_insensitive.rs
@@ -1,0 +1,184 @@
+// Copyright 2017 The Rust Project Developers. See the COPYRIGHT
+// file at the top-level directory of this distribution and at
+// http://rust-lang.org/COPYRIGHT.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+//! Timely dataflow runs on its own thread.
+
+use crate::facts::{AllFacts, Point};
+use crate::output::Output;
+use crate::output::timely_util::populate_args_for_differential_dataflow;
+use differential_dataflow::collection::Collection;
+use differential_dataflow::operators::*;
+use std::collections::BTreeSet;
+use std::mem;
+use std::sync::mpsc;
+use std::sync::Arc;
+use std::sync::Mutex;
+use timely;
+use timely::dataflow::operators::*;
+
+pub(super) fn compute(dump_enabled: bool, mut all_facts: AllFacts, workers: u32) -> Output {
+    // Declare that each universal region is live at every point.
+    let all_points: BTreeSet<Point> = all_facts
+        .cfg_edge
+        .iter()
+        .map(|&(p, _)| p)
+        .chain(all_facts.cfg_edge.iter().map(|&(_, q)| q))
+        .collect();
+
+    for &r in &all_facts.universal_region {
+        for &p in &all_points {
+            all_facts.region_live_at.push((r, p));
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


src/output/location_insensitive.rs
```diff
@@ -0,0 +1,182 @@
...
+                        .map(|(_r1, l, r2)| (r2, l));
+
+                    requires1.concat(&requires2).distinct_total()
+                });
+
+                // borrow_live_at(L, P) :-
+                //     requires(R, L), region_live_at(R, P)
+                let borrow_live_at = requires
+                    .join(&region_live_at)
+                    .map(|(_r, l, p)| (l, p));
```
nikomatsakis: you're missing a `distinct_total` here I believe

spastorino: You're right, probably a copy and paste from the previous part of the code where distinct_total is done in the concat line. Fixed!!!

PR #34, opened 2018-05-17
Make LocationInsensitive return potential errors

spastorino:
Another part of #29 to start discussion


nikomatsakis:
Also I see @lqd had a [few comments over on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/122657-wg-nll/subject/porting-to-datafrog/near/126913477)

src/output/mod.rs
```diff
@@ -85,6 +87,11 @@ impl Output {
                 intern,
                 &self.region_live_at,
             )?;
+            dump::dump_rows(
```
nikomatsakis: I think this should not be gated on `dump_enabled`

PR #36, opened 2018-05-18
🐸

lqd:
This switches from using @frankmcsherry’s differential dataflow to
using @frankmcsherry’s _bespoke_ datalog engine: datafrog.

99% of this commit was authored by the one and only @frankmcsherry.


src/output/datafrog_opt.rs
```diff
@@ -0,0 +1,294 @@
...
+        let subset_r1p = iteration.variable_indistinct("subset_r1p");
+        let subset_r2p = iteration.variable_indistinct("subset_r2p");
+        let subset_p = iteration.variable_indistinct("subset_p");
+
+        // temporaries as we perform a multi-way join, and more indices
+        let requires = iteration.variable::<(Region, Loan, Point)>("requires");
+        let requires_1 = iteration.variable("requires_1");
+        let requires_2 = iteration.variable("requires_2");
+        let requires_bp = iteration.variable("requires_bp");
+        let requires_rp = iteration.variable("requires_rp");
```
frankmcsherry: I apologize for not noticing this before, but each of the `requires_` variables can be `variable_indistinct()`. We just need `requires` itself (without a `_`) to be a `variable()` without `_indistinct()`. This has a solid perf benefit for me, taking the time from 9.43s to 7.99s.

lqd: Solid benefits for me as well, 10-15% 👍

src/output/naive.rs
```diff
@@ -8,258 +8,214 @@
...
+            //   cfg_edge(P, Q),
+            //   region_live_at(R1, Q),
+            //   region_live_at(R2, Q).
+
+            subset_1.from_join(&subset_p, &cfg_edge_p, |&_p, &(r1,r2), &q| ((r1,q),r2));
+            subset_2.from_join(&subset_1, &region_live_at, |&(r1,q),&r2,&()| ((r2,q),r1));
+            subset.from_join(&subset_2, &region_live_at, |&(r2,q),&r1,&()| (r1,r2,q));
+        }
+
+        subset_r1p.complete()
```
nikomatsakis: Just to check my understanding here: it seems to me that the `subset` variable is not needed. We *could* just use `subset_r1p` instead? (That is, nothing directly joins from `subset`; the only use of it is to be mapped into different arrangements, and we could do those maps from `subset_r1p` just as easily, right?)

(cc @frankmcsherry)

src/output/naive.rs
```diff
@@ -8,258 +8,214 @@
...
+        // some indexes
+        let requires_rp = iteration2.variable::<((Region, Point), Loan)>("requires_rp");
+        let requires_bp = iteration2.variable::<((Loan, Point), Region)>("requires_bp");
+
+        let requires_1 = iteration2.variable::<(Point, (Loan, Region))>("requires_1");
+        let requires_2 = iteration2.variable::<((Region, Point), Loan)>("requires_2");
+
+        // since we're using subset mapped ((r, p), r) we can use it directly out of iteration 1
+        let subset_r1p = iteration2.variable::<((Region, Point), Region)>("subset_r1p");
+        subset_r1p.insert(subset);
```
nikomatsakis: This "variable" is invariant, right? Do we really need to 'copy' the values from `subset` in here, or could we use `subset` directly?

(cc @frankmcsherry)

nikomatsakis: Well, let me answer my own question: it won't build if we don't, because `from_join` wants a `Variable` and not a `Relation`: presumably that could be modified in datafrog? (Or is this `insert` very cheap anyway?)

nikomatsakis: (I realize in this specific case we ought to collapse the iteration anyway.)

nikomatsakis: (this is also relevant, it seems to the "redundantly computed index" entries we see below)

src/output/naive.rs
nikomatsakis: Answer: yes, I checked elsewhere, and even got a slight win from doing so (in datafrog-opt).

nikomatsakis:
Final measurements:

| Version | Workers | Compile Time | Run Time |
| --- | --- | --- | --- |
| :timer_clock:  | 1 | 210s | 14.191s |
| :timer_clock:  | 2 | 210s | 11.035s |
| :timer_clock:  | 4 | 210s | 8.898s |
| :timer_clock:  | 8 | 210s | 7.953s |
| :frog:  | 1 | 11s | 6.843s |

Amazing! Great work @frankmcsherry and @lqd!

src/output/naive.rs
frankmcsherry: Inserting a sorted relation should be pretty cheap. It will still do linear work, confirming that none of `subset` exists in the empty list of prior tuples, which we could totally optimize. But it shouldn't be too expensive either way.

More generally, there a bunch of cases where the difference between `Variable` and `Relation` and between `Key` and `(Key, ())` bite us a little. These can be picked off as appropriate; just some repetition in the library code in the interest of squeaking out some perf. :)

frankmcsherry: Oh, my mistake. It will not even do the linear work, I think, on account of `self.stable` should be empty. I could be wrong about that, and we could work through exactly what happens when you install a relation. Not a great amount of care was used putting the wrappers in place. :)

PR #39, opened 2018-05-19
Simplify naive datafrog analysis

### LINKED ISSUE #37: Simplify the "Naive" datafrog analysis
The Naive version of the datafrog analysis is a bit more complex than it needs to be, reflecting a structure that actually doesn't exist in the datalog program itself:

- it uses 3 datafrog `Iterations` when only one is effectively needed (as can be seen in the `DatafrogOpt` analysis variant, which also only uses one iteration)
- as a result, there are some unnecessary copies of relations/indices data (those are marked in the code, for example [this one](https://github.com/rust-lang-nursery/polonius/blob/master/src/output/naive.rs#L59)) since more than one `Iteration` uses it.

It would be good remove those, resulting in the most obvious datalog-to-datafrog conversion (and faster as well).

The results of this analysis can be directly compared to the `DatafrogOpt` analysis results, so it would be easy to avoid mistakes as well.

KiChjang:
Removes all unnecessary iterations and data copies.

Fixes #37.


nikomatsakis:
Nice — I wanted to double check that we still produce same tuples before I merged, but this basically looks right to me.

lqd:
Probably not a big deal, but this does change the output of `-v` and `--stats` from the previous datafrog and timely versions.

It'd be nice to keep the comments up-to-date : 
- removing the indices makes them not `// redundantly computed index` anymore
- the [note at the top](https://github.com/rust-lang-nursery/polonius/blob/2b7ce59b0f7cafac11ad64e6513fea0790285b5d/src/output/naive.rs#L21-L23) can now be removed

KiChjang:
@nikomatsakis I've tried diffing the output against the dataopt strategy -- there are no differences at all.

Issue #26, opened 2018-05-14
friendlier front end


nikomatsakis:
Right now, the "input" to this program is a bunch of facts you could never generate by hand. We should add some sort of front-end more like chalk, that takes a human-writable format -- perhaps similar to MIR? And generates facts from it.  This would allow us to do more unit testing.

It'd probably be good then to factor out the "core algorithm" into a separate crate, so that it can eventually be shared by rustc and this crate.

I'd be open to suggestions for what the input format should look like =)

nikomatsakis:
I've been debating about this. I think my *ideal* would be that we take in MIR and generate facts, but that would require reproducing a lot of the compiler. Obviously the simplest format is the one we already have: just give me the raw facts. So the question is, which facts should we "derive" from  a nicer input and which can the user manually specify?

It seems to me that by far the most tedious to derive are `cfg_edge` and `region_live_at`. So pehaps we should focus on those, and have the user manually specify the `borrow_region`, `outlives`, and `invalidates` themselves. We can then (over time) transition to the rest.

I envision something like this. First, a program consists of a `universal_regions` declaration and a set of basic blocks:

```
universal_regions { 'a, 'b, 'c }
block B0 {
    statement0; // series of `;`-separated statements, defined below
    statement1;
    goto B1, B2, B3; // create edges between basic blocks
}
```

This is kind of a grammar like this (using a LALRPOP-like format):

```
Input = UniversalRegions BlockDefn*;
UniversalRegions = "universal_regions" "{" Comma<Region> "}";
BlockDefn = "block" Block "{" Statement* Goto? "}"
Goto = "goto" Comma<Block> ";"
```


Each statement would be a comma-separate list of facts that occur on that line, or possibly a `use`. So something like:

```
use('a), outlives('a: 'b), borrow_region_at('b, 'c)
```

which might fit this grammar:

```
Statement = { Fact, Use };
Fact = {
  "outlives" "(" Region ":" Region ")",
  "borrow_region_at" "(" Region "," Loan ")",
  "invalidates" "(" Loan ")",
};
Use = "use" "(" Region ")" 
```

And of course we need the base strings:

```
Block = r"B\w+"; // or whatever
Region = r"'\w+"; // ...
Loan = r"L\w+"; // ...
```

We would then convert these into tuples of `AllFacts` and run the computation.

PR #45, opened 2018-05-23
Implement friendly front-end format

lqd:
Implement the friendlier front-end format described in issue #26 — destined to help create isolated tests without needing rustc NLL facts, but still generating the same data structure.

Here's how it looks:
```
// program description
universal_regions { 'a, 'b, 'c }

// block description
block B0 {
    // effects are by default emitted on the Mid point
    invalidates(L0);

    // but we can still emit some at the Start point if needed
    // by separating the Start point effects from the Mid point effects with a "/"
    invalidates(L1) / kill(L2);

    goto B1;
}
block B1 {
    use('a), outlives('a: 'b), borrow_region_at('b, L1);
}
```

### DIFF
--- a/src/program.rs
+++ b/src/program.rs
@@ -1,0 +1,335 @@
+use std::collections::BTreeSet;
+
+use polonius_parser::{
+    ir::{Effect, Fact}, parse_input,
+};
+
+use crate::facts::{AllFacts, Loan, Point, Region};
+use crate::intern::InternerTables;
+
+/// A structure to hold and deduplicate facts
+#[derive(Default)]
+struct Facts {
+    borrow_region: BTreeSet<(Region, Loan, Point)>,
+    universal_region: BTreeSet<Region>,
+    cfg_edge: BTreeSet<(Point, Point)>,
+    killed: BTreeSet<(Loan, Point)>,
+    outlives: BTreeSet<(Region, Region, Point)>,
+    region_live_at: BTreeSet<(Region, Point)>,
+    invalidates: BTreeSet<(Point, Loan)>,
+}
+
+impl From<Facts> for AllFacts {
+    fn from(facts: Facts) -> Self {
+        Self {
+            borrow_region: facts.borrow_region.into_iter().collect(),
+            universal_region: facts.universal_region.into_iter().collect(),
+            cfg_edge: facts.cfg_edge.into_iter().collect(),
+            killed: facts.killed.into_iter().collect(),
+            outlives: facts.outlives.into_iter().collect(),
+            region_live_at: facts.region_live_at.into_iter().collect(),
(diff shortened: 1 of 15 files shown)


src/program.rs
```diff
@@ -0,0 +1,315 @@
...
+                    let to = tables.points.intern(&to);
+
+                    facts.cfg_edge.insert((from, to));
+                }
+            }
+
+            // the most common statement effects: mid point effects
+            for effect in &statement.effects {
+                match effect {
+                    // facts: region_live_at(Region, Point)
```
nikomatsakis: Hmm, so for `region_live_at`, we have to do more — we need to actually compute liveness I think, but I realize that to do that we need to also have `def('a)`, of course (and maybe we should make both `def` and `use` accept multiple lifetimes). We could also try to capture the idea of variables with types, though that is creeping towards MIR and I think I had hoped to do that as a second pass.

In other words, as an example:

```
bb0 {
  def('a); // statement 0
  ; // statement 1
  goto bb1;
}

bb1 {
  ; // statement 0
  use('a); // statement 1
}
```

This ought to make `'a` live at all the following points: 

- start(bb0/1), 
- mid(bb0/1), 
- start(bb1/0), 
- mid(bb1/0), 
- start(bb1/1), 
- mid(bb1/1).

The idea would be: some variable `X` whose type includes `'a` is used at mid(bb1/1). So, `'a` is live at that point, because the current value of `X` matters -- this flows backwards until we reach `mid(bb0/0)`, at which point `X` was assigned, so its current value doesn't matter.

nikomatsakis: We can of course use datafrog to do this computation =) you can imagine us emitting `use` and `def` facts and setting up a rule:

```rust
region_live_at(R, P) :- use(R, P).
region_live_at(R, P) :-
  use(R, Q),
  cfg_edge(P, Q),
  !def(R, P).
```

src/program.rs
```diff
@@ -0,0 +1,315 @@
...
+        }
+
+        // facts: invalidates(Point, Loan)
+        Effect::Invalidates { ref loan } => {
+            let loan = tables.loans.intern(loan);
+            // invalides: a loan can be invalidated on both Start and Mid points
+            facts.invalidates.insert((point, loan));
+        }
+
+        _ => panic!("Unexpected effect: {:?}", effect),
```
nikomatsakis: I would rather factor `Effect` into nested enums so that this case never happens.

src/program.rs
nikomatsakis: We also could potentially refactor the compiler to emit `use` facts and not `region_live_at` facts.

lqd:
Alright I'll fix this and reopen this later tonight

PR #47, opened 2018-05-24
make the output code generic over region, loan, point

nikomatsakis:
Following up on @bobtwinkles PR, because I couldn't stop myself.


nikomatsakis:
or r? @lqd

lqd:
This refactor looks nice, r+ 👍 

potential nit: since you removed the stats computation, we could also remove `cli::Opt::stats` as well.

PR #52, opened 2018-05-25
Move Output to polonius-engine

spastorino:
This is what we did with @qmx


polonius-engine/Cargo.toml
```diff
@@ -9,3 +9,5 @@ readme = "README.md"
 keywords = ["compiler", "borrowck", "datalog"]
 
 [dependencies]
+datafrog = "0.1.0"
+fxhash = "0.2.1"
```
nikomatsakis: can you use `rustc_hash` instead? It's a new crate I added recently https://crates.io/crates/rustc-hash

just change to `rustc-hash = 1.0.0`

PR #67, opened 2018-05-30
Add facts from issue-31567, where Naive and Opt diverge

lqd:
A simple case where the Naive and DatafrogOpt variants compute different `borrow_live_at` tuples.

The Naive variant computes 2 additional loans that Opt doesn't:
```
"Mid(bb0[4])"    "bw1"
"Mid(bb0[8])"    "bw3"
```

The 2 missing borrows we'd expect to see are from the `borrow_region` input tuples (making them `requires` tuples) but are not in the `region_live_at` input -- thus making the last rule `borrow_live_at(B, P) :- requires(R, B, P), region_live_at(R, P)` ignore them.

The function where Naive and Opt differ is `get_dangling`, and e.g. the `-a compare` command option will fail (however, this command is comparing the errors, and the differing point before the ones computing errors)

Filing as the start of the "-Zpolonius" differences analysis, and as a future regression test.


lqd:
Closing, as I've completed the "-Z polonius" difference analysis between Naive and Opt, and will add those as unit tests in PR #69 instead.

PR #74, opened 2018-07-13
Remove `termination_trait_test` feature attribute

lqd:
The feature is stable since 1.27.0 and no longers require enabling — so it triggers a warning if we keep it.


lqd:
closing in favor of #77

Issue #62, opened 2018-05-29
The `subset` relation contains regions subsets of themselves


lqd:
We might need to investigate the symmetric `subset` tuples produced, if we can remove them, and if this has performance implications: in the clap benchmark, we can see 480K regions subsets of themselves.

To see these tuples, we can get debug output like the following (for example, right before filling the output tuples in `-v / dump_enabled` mode):
```rust
println!("subset ({}) symmetries: {:?}", subset.len(), subset.iter().filter(|&(r1, r2, _)| r1 == r2).count());
```

Outputting these results over the clap dataset:
1) ```subset (7531526) symmetries: 794833``` for the Naive variant
2) ```subset (2714121) symmetries: 481294``` for the DatafrogOpt variant

PR #78, opened 2018-09-21
Remove symmetries in the `subset` relation

lqd:
On the clap dataset, there are:
- Naive variant: subset (7 531 526) symmetries: 794 833
Removing the symmetries results: from 36.7s to 33.2s (non-scientific benchmarks, best time)

- DatafrogOpt variant: subset (2 714 121) symmetries: 481 294
Removing the symmetries results: from 5.89s to 5.02s (non-scientific benchmarks, best time)

This could be considered as fixing #62.

r? @nikomatsakis


nikomatsakis:
Hmm, hacky, but ... ok I guess.

lqd:
For the record, my preferred 3rd way of doing this would be to add methods with `filter_map` capability to datafrog's map, join, antijoin, and treefrog joins.

PR #84, opened 2018-11-27
Dead can reach bug fix

nikomatsakis:
The `borrow_live_at` relation as defined today is correct so long as
the "requires" originates from some previous statement.  But it is
overlooking one case where a borrow can be live within a single point
`P`:

If you have a `borrow_live_at(R, P, B)` at some point P where:

- `R` is dead on entry to P
- but `R <=* R2`
- and `R2` is live

then we don't handle this at all. So add some logic to detect this
case. Also do a bunch of other random cleanup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ particularly egregious case where we currently perform poorly, and you
 can test it against it like so:
 
 ```bash
-> cargo +nightly run --release -- inputs/clap-rs/app-parser-{{impl}}-add_defaults/ | head
+> cargo +nightly run --release -- inputs/clap-rs/app-parser-{{impl}}-add_defaults/
     Finished release [optimized] target(s) in 0.05 secs
      Running `target/release/borrow-check 'inputs/clap-rs/app-parser-{{impl}}-add_defaults/'`
 --------------------------------------------------
--- a/README.md
+++ b/README.md
@@ -35,23 +35,29 @@ with input and output tuples at each point, do:
 cargo +nightly run --release -- --graphviz_file=graph.dot inputs/issue-47680/nll-facts/main
 ```
 
-### Want to see something slow?
+### Want to run the code?
 
 One of the goals with this repo is to experiment and compare different
-implementations of the same algorithm. The repo includes one
-particularly egregious case where we currently perform poorly, and you
-can test it against it like so:
+implementations of the same algorithm. You can run the analysis by using `cargo run`
+and you can choose the analysis with `-a`. So for example to run against an example
+extract from clap, you might do:
 
 ```bash
-> cargo +nightly run --release -- inputs/clap-rs/app-parser-{{impl}}-add_defaults/
+> cargo +nightly run --release -- -a DatafrogOpt inputs/clap-rs/app-parser-{{impl}}-add_defaults/
     Finished release [optimized] target(s) in 0.05 secs
      Running `target/release/borrow-check 'inputs/clap-rs/app-parser-{{impl}}-add_defaults/'`
 --------------------------------------------------
 Directory: inputs/clap-rs/app-parser-{{impl}}-add_defaults/
-Time: 113.316s
+Time: 3.856s
 ```
 
-(You can see it is pretty dang slow on my machine!)
+You could also try `-a Naive` to get the naive rules (more readable, slower)
+or `-a LocationInsensitive` to use a location insensitive analysis.
+
+By default, `cargo run` just prints timing. If you also want to see
+the results, try `--show-tuples` (which will show errors) and maybe
+`-v` (to show more intermediate computations). You can supply `--help`
+to get more docs.
 
 ### How to generate your own inputs
 
--- a/README.md
+++ b/README.md
@@ -3,6 +3,14 @@ This is a core library that models the borrow check. It implements the analysis
 
 [post]: http://smallcultfollowing.com/babysteps/blog/2018/04/27/an-alias-based-formulation-of-the-borrow-checker/
 
+### Why the name "Polonius"?
+
+The name comes from the famous quote ["Neither borrower nor lender
+be"][nblnb], which comes from the character Polonius in Shakespeare's
+*Hamlet*.
+
+[nblnb]: https://literarydevices.net/neither-a-borrower-nor-a-lender-be/
+
 ### How to use
 
 First off, you must use the **nightly** channel. To build, do something like this:
--- a/README.md
+++ b/README.md
@@ -11,38 +11,6 @@ be"][nblnb], which comes from the character Polonius in Shakespeare's
 
 [nblnb]: https://literarydevices.net/neither-a-borrower-nor-a-lender-be/
 
-### How to use
-
-First off, you must use the **nightly** channel. To build, do something like this:
-
-```bash
-cargo +nightly build --release
-```
-
-You can try it on one our input tests like so:
-
-```bash
-cargo +nightly run --release -- inputs/issue-47680/nll-facts/main
-```
-
-This will generate a bunch of output tuples:
-
-```
-# borrow_live_at
-
-"Mid(bb3[2])"   "bw0"
-"Mid(bb3[2])"   "bw2"
-"Mid(bb10[2])"  "bw0"
-...
-```
-
-To generate a [graphviz](https://www.graphviz.org) file with the CFG enriched
-with input and output tuples at each point, do:
-
-```bash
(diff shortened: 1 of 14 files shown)


src/test_util.rs
```diff
@@ -0,0 +1,46 @@
+#![cfg(test)]
+
+use std::fmt::Debug;
+
+/// Test that two values are equal, with a better error than `assert_eq`
+pub fn assert_equal<A>(expected_value: &A, actual_value: &A)
```
lqd: Diff output: super useful with our hard to read results 👍. No need to change anything since you've done the work, but if we ever want _fancy color output_, [pretty-assertions](https://github.com/colin-kiegel/rust-pretty-assertions) seemed to work nicely.

polonius-engine/src/output/datafrog_opt.rs
```diff
@@ -91,7 +91,6 @@ pub(super) fn compute<Region: Atom, Loan: Atom, Point: Atom>(
 
         let dying_can_reach_live =
             iteration.variable::<((Region, Point, Point), Region)>("dying_can_reach_live");
-        let dying_can_reach_live_r1pq = iteration.variable_indistinct("dying_can_reach_live_r1pq");
```
lqd: Sorry I should have left a comment before that I had noticed this variable, and when I tried removing it I had the same experience as [yours](https://github.com/rust-lang-nursery/polonius/blob/master/polonius-engine/src/output/datafrog_opt.rs#L89-L90): it was slightly slower on clap here as well. Probably just noise.

polonius-engine/src/output/datafrog_opt.rs
```diff
@@ -181,37 +181,42 @@ pub(super) fn compute<Region: Atom, Loan: Atom, Point: Atom>(
...
             // can reach.
-            dead_can_reach_origins.from_map(&live_to_dead_regions, |&(_r1, r2, p, q)| ((r2, p), q));
-            dead_can_reach_origins.from_map(&dead_region_requires, |&((r, p, q), _b)| ((r, p), q));
-
-            // .decl dead_can_reach(R1, R2, P, Q)
+            //
+            // dying_can_reach_origins(R2, P, Q) :-
+            //   live_to_dead_regions(_, R2, P, Q).
+            // dying_can_reach_origins(R, P, Q) :-
+            //   dying_region_requires_2(R, P, Q, B).
```
lqd: nit: `dying_region_requires_2(R, P, Q, _)`

polonius-engine/src/output/datafrog_opt.rs
```diff
@@ -88,19 +91,29 @@ pub(super) fn compute<Region: Atom, Loan: Atom, Point: Atom>(
         let dying_can_reach_live =
             iteration.variable::<((Region, Point, Point), Region)>("dying_can_reach_live");
 
+        let dead_borrow_region_can_reach_root =
+            iteration.variable::<((Region, Point), Loan)>("dead_borrow_region_can_reach_root");
+        let dead_borrow_region_can_reach_dead =
+            iteration.variable::<((Region, Point), Loan)>("dead_borrow_region_can_reach_dead");
+        let dead_borrow_region_can_reach_dead_1 =
+            iteration.variable("dead_borrow_region_can_reach_dead_1");
```
lqd: Being an intermediary-step variable, `dead_borrow_region_can_reach_dead_1` could work as a "distinct-less" Variable (`variable_indistinct`) right ?

polonius-engine/src/output/datafrog_opt.rs
```diff
@@ -308,6 +321,39 @@ pub(super) fn compute<Region: Atom, Loan: Atom, Point: Atom>(
...
+                &subset_r1p,
+                |&(_r1, p), &b, &r2| ((r2, p), b),
+            );
+            dead_borrow_region_can_reach_dead.from_antijoin(
+                &dead_borrow_region_can_reach_dead_1,
+                &region_live_at_rel,
+                |&(r2, p), &b| ((r2, p), b),
+            );
+
+            // dead_borrow_region_can_reach_live(B, P) :-
```
lqd: stray comment I think

polonius-engine/src/output/datafrog_opt.rs
```diff
@@ -316,6 +362,16 @@ pub(super) fn compute<Region: Atom, Loan: Atom, Point: Atom>(
                 ((b, p), ())
             });
 
+            // borrow_live_at(B, P) :-
+            //   dead_borrow_region_can_reach_dead(R1, B, P),
+            //   subset(R1, R2, P),
```
lqd: I believe this is a copy-paste typo since we're not using `subset` here (subsets being handled in `dead_borrow_region_can_reach_dead` already) ? 
So the following line may be `region_live_at(R1, P)`, and the closure argument `_r1` — but that would be almost too nitpicky (_but_ still not worthy of a "pacify the mercilous $reviewer" commit, I hope)

README.md
```diff
@@ -35,23 +35,29 @@ with input and output tuples at each point, do:
...
     Finished release [optimized] target(s) in 0.05 secs
      Running `target/release/borrow-check 'inputs/clap-rs/app-parser-{{impl}}-add_defaults/'`
 --------------------------------------------------
 Directory: inputs/clap-rs/app-parser-{{impl}}-add_defaults/
-Time: 113.316s
+Time: 3.856s
 ```
 
-(You can see it is pretty dang slow on my machine!)
+You could also try `-a Naive` to get the naive rules (more readable, slower)
```
lqd: would it be useful to mention that these rules are in the fact the ones described in the blog post ?

src/cli.rs
```diff
@@ -18,15 +18,15 @@ pub struct Opt {
...
+    #[structopt(long = "show-tuples", help = "Show output tuples on stdout")]
     show_tuples: bool,
-    #[structopt(long = "skip-timing")]
+    #[structopt(long = "skip-timing", help = "Do not display timing results")]
     skip_timing: bool,
-    #[structopt(short = "v")]
+    #[structopt(short = "v", long = "verbose", help = "Show intermediate output tuples and not just errors")]
     verbose: bool,
-    #[structopt(long = "graphviz_file")]
+    #[structopt(long = "graphviz_file", help = "Generate a graphviz file to visualize the input")]
```
lqd: nit: IIRC the graphviz file contains some output tuples in addition to the input facts, so maybe "to help visualize the computation" ?

polonius-engine/src/output/datafrog_opt.rs
nikomatsakis: Yeah. I tried measuring and didn't really notice anything reliable. It's hard to imagine though that it's *better* to have extra variables, right?

src/test_util.rs
nikomatsakis: I think I tried `pretty-assertions` and found that it was not giving me reasonable diffs (iirc, it was not handling `\n` correctly, so I just got all `-` rows). Perhaps I was doing something silly though.

polonius-engine/src/output/datafrog_opt.rs
nikomatsakis: True.

polonius-engine/src/output/datafrog_opt.rs
nikomatsakis: Well, it depends. I am using the **intermediate variable** `dead_borrow_region_can_reach_dead_1`, which is equal to `dead_borrow_region_can_reach_dead(R1, B, P), subset(R1, R2, P)`. So yes you don't see `subset` below but the effect is still true to the datalog as written. Maybe I should comment *that*?

polonius-engine/src/output/datafrog_opt.rs
nikomatsakis: Ah yes that was an older attempt

src/test_util.rs
nikomatsakis: Oh no, I was trying some other crate. That *does* look nice. However, maybe i'll still with this, because I was hoping to change from using `Debug` output to something else that would let us give names to the values (not just the interned indices)

polonius-engine/src/output/datafrog_opt.rs
lqd: Yeah and in any case we always document the datalog semantics and not the literal datafrog syntax as I was wrongly mentioning. The _1 was indeed subtle but regardless: you're absolutely right and my comment was misguided. 

The new comment is perfectly clear.

src/test_util.rs
lqd: Yes, the `debug_with` plan ! Sounds good.

PR #88, opened 2018-12-05
adopt leapfrog

nikomatsakis:
Depends on https://github.com/rust-lang-nursery/datafrog/pull/16 being published first =)

### DIFF
--- a/polonius-engine/src/output/datafrog_opt.rs
+++ b/polonius-engine/src/output/datafrog_opt.rs
@@ -64,8 +64,7 @@ pub(super) fn compute<Region: Atom, Loan: Atom, Point: Atom>(
         // variables, indices for the computation rules, and temporaries for the multi-way joins
         let subset_r1p = iteration.variable::<((Region, Point), Region)>("subset_r1p");
 
-        let requires_bp = iteration.variable::<((Loan, Point), Region)>("requires_bp");
-        let requires_rp = iteration.variable_indistinct("requires_rp");
+        let requires_rp = iteration.variable::<((Region, Point), Loan)>("requires_rp");
 
         let borrow_live_at = iteration.variable::<((Loan, Point), ())>("borrow_live_at");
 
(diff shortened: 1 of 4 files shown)


polonius-engine/src/output/datafrog_opt.rs
```diff
@@ -47,7 +47,9 @@ pub(super) fn compute<Region: Atom, Loan: Atom, Point: Atom>(
 
         // static inputs
         let cfg_edge = iteration.variable::<(Point, Point)>("cfg_edge");
-        let killed = all_facts.killed.into();
+        let cfg_edge_rel = Relation::from(all_facts.cfg_edge.iter().map(|&(p, q)| (p, q)));
```
lqd: For posterity's sake (as you know this): I don't think we need the CFG edges as a `Variable` anymore with the leapjoin

polonius-engine/src/output/datafrog_opt.rs
```diff
@@ -62,28 +64,20 @@ pub(super) fn compute<Region: Atom, Loan: Atom, Point: Atom>(
 
         // variables, indices for the computation rules, and temporaries for the multi-way joins
         let subset = iteration.variable::<(Region, Region, Point)>("subset");
-        let subset_1 = iteration.variable_indistinct("subset_1");
-        let subset_2 = iteration.variable_indistinct("subset_2");
         let subset_r1p = iteration.variable_indistinct("subset_r1p");
         let subset_p = iteration.variable_indistinct("subset_p");
```
lqd: For posterity's sake: I think we can also get by with only one `subset` variable

polonius-engine/src/output/datafrog_opt.rs
```diff
@@ -62,28 +64,20 @@ pub(super) fn compute<Region: Atom, Loan: Atom, Point: Atom>(
...
 
         let requires = iteration.variable::<(Region, Loan, Point)>("requires");
-        let requires_1 = iteration.variable_indistinct("requires_1");
-        let requires_2 = iteration.variable_indistinct("requires_2");
         let requires_bp = iteration.variable_indistinct("requires_bp");
         let requires_rp = iteration.variable_indistinct("requires_rp");
```
lqd: For posterity's sake: I think we can also have only one `requires` variable

PR #90, opened 2018-12-06
Port the LocationInsensitive variant to leapjoin

lqd:
On the `clap` dataset:
- before: around 370ms
- after: around 130ms

r? @nikomatsakis 

A couple remarks:
1) This removes materializing the `borrow_live_at` tuples to take full advantage of the leapjoin. While it would be premature to remove this relation from DatafrogOpt at this point, it seemed low-risk for the LocationInsensitive variant. If need be, we can keep probably keep it. What do you think ?

2) While the other join is between variables, one of them is static and could be ported to leapjoin, however doing so was 7% slower on `clap` so I'm not sure it's worth doing ?

3) I've also removed the verbose output of the `invalidates` input. I'm not sure why it was there, it's the only variant which does that, and it didn't seem that useful.

### DIFF
--- a/polonius-engine/src/output/location_insensitive.rs
+++ b/polonius-engine/src/output/location_insensitive.rs
@@ -13,7 +13,7 @@ use std::time::Instant;
 
 use crate::output::Output;
 
-use datafrog::{Iteration, Relation};
+use datafrog::{Iteration, Relation, RelationLeaper};
 use facts::{AllFacts, Atom};
 
 pub(super) fn compute<Region: Atom, Loan: Atom, Point: Atom>(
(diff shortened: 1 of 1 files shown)


polonius-engine/src/output/location_insensitive.rs
```diff
@@ -64,22 +66,28 @@ pub(super) fn compute<Region: Atom, Loan: Atom, Point: Atom>(
...
-        invalidates.insert(Relation::from(
-            all_facts.invalidates.iter().map(|&(p, b)| ((b, p), ())),
-        ));
-
         // .. and then start iterating rules!
         while iteration.changed() {
             // requires(R2, B) :- requires(R1, B), subset(R1, R2).
+            //
+            // Note: Since `subset` is effectively a static input, this join can be ported to
+            // a leapjoin. Doing so, however, was 7% slower on `clap`.
```
nikomatsakis: Makes sense: I imagine that a plain "join" is more efficient if there are exactly two inputs.

PR #91, opened 2018-12-06
Deduplicate dependencies

lqd:
Making polonius build a tiny bit faster (from scratch). There are still some duplicate crates (e.g. 3 `regex-syntax`) for now.

`grep "name = " Cargo.lock | awk '{ print $3 }' | sort | uniq -c | grep -v "1 \""`
Before:
```
2 "atty"
2 "bitflags"
2 "proc-macro2"
2 "quote"
2 "rand"
2 "regex"
3 "regex-syntax"
2 "strsim"
2 "syn"
2 "winapi"
```

After:
```
2 "atty"
2 "error-chain"
2 "rand"
2 "rand_core"
2 "regex"
3 "regex-syntax"
2 "strsim"
2 "winapi"
```

For the others, and especially the triple regex-syntax, I'll consult with the dependency de-duplication resident expert @eijebong :)

-----
Also, note that this updates the lockfile to include the `polonius-engine` 0.6.1 release, which I think was missing from #89.

r? @nikomatsakis

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -15,11 +15,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "arrayref"
-version = "0.3.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-
-[[package]]
 name = "ascii-canvas"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


lqd:
Pushed a couple commits:
- Updated `lalrpop`to 0.16 (which itself also builds faster)
- Updated `assert_cli` to 0.6
- Removed an unused crate (`histo`).

The number of duplicate crates has gone down as expected:
```
2 "rand_core"
2 "winapi"
```

PR #93, opened 2018-12-06
Error generation: add tests and fix Naive variant

lqd:
This PR:
- adds the facts from rustc's `polonius_smoke_test`: it's a dataset containing both positive and negative cases for polonius, making it easy to test error generation both in unit tests and manually.
- adds missing `invalidates` inputs to the `Naive` variant, to successfully generate errors.

r? @nikomatsakis


lqd:
Maybe I should have added the successful tests to the `tests!` list ? 

(I hadn't done that in order to add assertions about the `LocationInsensitive` variant for those 2 tests, but maybe they're not that important — and if so, adding them using the macro instead would reduce code duplication)

src/test.rs
```diff
@@ -207,3 +208,76 @@ fn borrowed_local_error() {
     let facts = parse_from_program(program, &mut tables).expect("Parsing failure");
     test_facts(&facts, Algorithm::OPTIMIZED);
 }
+
+#[test]
+fn smoke_test_errors() {
```
nikomatsakis: ugh. good to test, but we gotta make this more concise.

polonius-engine/src/output/naive.rs
```diff
@@ -79,6 +79,9 @@ pub(super) fn compute<Region: Atom, Loan: Atom, Point: Atom>(
         // load initial facts.
         subset.insert(all_facts.outlives.into());
         requires.insert(all_facts.borrow_region.into());
+        invalidates.insert(Relation::from(
+            all_facts.invalidates.iter().map(|&(p, b)| ((b, p), ())),
+        ));
```
nikomatsakis: ....huh. Whoops.

PR #96, opened 2018-12-27
extend tests to cover the insensitive analysis

nikomatsakis:
r? @lqd

### DIFF
--- a/src/test.rs
+++ b/src/test.rs
@@ -16,8 +16,29 @@ fn test_facts(all_facts: &AllFacts, algorithms: &[Algorithm]) {
     // If the insensitive analysis concludes no errors, then naive
     // should also.
     let insensitive = Output::compute(all_facts, Algorithm::LocationInsensitive, false);
-    if insensitive.errors.is_empty() {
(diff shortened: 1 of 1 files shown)


src/test.rs
```diff
@@ -12,6 +12,15 @@ use std::path::Path;
 
 fn test_facts(all_facts: &AllFacts, algorithms: &[Algorithm]) {
     let naive = Output::compute(all_facts, Algorithm::Naive, true);
+
+    // If the insensitive analysis concludes no errors, then naive
+    // should also.
+    let insensitive = Output::compute(all_facts, Algorithm::LocationInsensitive, false);
+    if insensitive.errors.is_empty() {
+        assert_equal(&naive.errors, &insensitive.errors);
```
nikomatsakis: maybe we could strengthen this to a subset test

lqd: it could be interesting indeed

lqd:
(nice of github to show the comment more than once)

nikomatsakis:
Rebased, added subset.

PR #99, opened 2019-01-18
[WIP] Begin implementing some illegal subset relations errors

lqd:
Opening for discussion/guidance: especially around how we expect errors to work, how to test these errors cleanly if this is an acceptable way of outputting them, how to have better organized tests, etc. 

Implements the rules from the [post](http://smallcultfollowing.com/babysteps/blog/2019/01/17/polonius-and-region-errors/) to check whether named lifetimes have correct relations.

I've manually crafted the `known_subset` facts for the datasets (until we regenerate them with rustc when we can) so they may be wrong; for the tiny implied bounds one, I feel it should be similar to the regular `valid_subset` test, but I'm not sure about it.

As mentioned earlier: this doesn't check the subset errors in the unit tests yet, I only eyeballed them running our datasets, and will add those tests when we know/decide how to best.


nikomatsakis:
(We would clearly want some more tests)

lqd:
This WIP work is using technique number 1 out of 3 possible implementations, and can only be applied to the Naive variant since it uses the complete subset TC. 

This is not likely the way forward, probably rather "placeholder loans" will be the way to go. So I'm closing until this PR until I reimplement these errors with the expected technique.

PR #102, opened 2019-03-11
Add a naïve hybrid algorithm

### LINKED ISSUE #100: hybrid algorithm
As discussed in [today's meeting](https://rust-lang.zulipchat.com/#narrow/stream/186049-t-compiler.2Fwg-polonius/topic/meeting.202019.2E03.2E07), a simple starter issue for getting more understanding of how polonius works is to create a "hybrid" algorithm that first tries the location-insensitive analysis and -- if there are errors -- falls back to datafrog-opt.

albins:
Closes #100.

### DIFF
--- a/polonius-engine/src/output/hybrid.rs
+++ b/polonius-engine/src/output/hybrid.rs
@@ -20,7 +20,7 @@ pub(super) fn compute<Region: Atom, Loan: Atom, Point: Atom>(
     dump_enabled: bool,
     all_facts: AllFacts<Region, Loan, Point>,
 ) -> Output<Region, Loan, Point> {
-    let lins_output = location_insensitive::compute(dump_enabled, all_facts.clone());
+    let lins_output = location_insensitive::compute(dump_enabled, &all_facts);
     if lins_output.errors.is_empty() {
         lins_output
     } else {
--- a/polonius-engine/src/output/hybrid.rs
+++ b/polonius-engine/src/output/hybrid.rs
@@ -1,0 +1,29 @@
+// Copyright 2019 The Rust Project Developers. See the COPYRIGHT
+// file at the top-level directory of this distribution and at
+// http://rust-lang.org/COPYRIGHT.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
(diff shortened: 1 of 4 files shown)


polonius-engine/src/output/hybrid.rs
```diff
@@ -0,0 +1,28 @@
...
+use crate::output::datafrog_opt;
+use crate::output::location_insensitive;
+use crate::output::Output;
+use facts::{AllFacts, Atom};
+
+pub(super) fn compute<Region: Atom, Loan: Atom, Point: Atom>(
+    dump_enabled: bool,
+    all_facts: AllFacts<Region, Loan, Point>,
+) -> Output<Region, Loan, Point> {
+    let lins_output = location_insensitive::compute(dump_enabled, all_facts.clone());
```
nikomatsakis: I wonder if we should modify `location_insensitive::compute` so that it takes `&all_facts` instead, so that we don't have to clone here. From looking at the code it seems like it doesn't clone a lot of the things inside, but rather copies out itself, so it'd be overall more efficient.

albins: I’ll have a look at what happens if I change it and report back!

albins: Ok; done! In some shape or form anyway.

polonius-engine/src/output/location_insensitive.rs
```diff
@@ -27,12 +27,11 @@ pub(super) fn compute<Region: Atom, Loan: Atom, Point: Atom>(
         .chain(all_facts.cfg_edge.iter().map(|&(_, q)| q))
         .collect();
 
-    all_facts
-        .region_live_at
-        .reserve(all_facts.universal_region.len() * all_points.len());
+    let mut my_region_live_at = all_facts.region_live_at.clone();
+    my_region_live_at.reserve(all_facts.universal_region.len() * all_points.len());
```
nikomatsakis: Nit: I'd probably drop the `my_` prefix

albins: Fixed this in https://github.com/rust-lang/polonius/pull/102/commits/833ba373baff68404b5c52d0fcf34b6bbea1ca50, which also motivates why I think this is ok style-wise in the commit message.

PR #103, opened 2019-03-19
Version-bump polonius-engine to 0.7.0

albins:


albins:
> Can you update the [RELEASES.md](https://github.com/rust-lang/polonius/blob/master/RELEASES.md) file with a quick log of what has happened? I usually skim the `git log --merge-only` from the tag from the most recent release (hopefully I remembered to make one)

Yup, done!

Issue #104, opened 2019-03-21
incorporate liveness computation into this crate


nikomatsakis:
This is part of a general move to incorporate more of the borrow check computations into polonius itself. The overall goal here is that rustc will supply polonius with facts that say "this variable is defined here, this variable is used there" as well as which regions appear within the variable's type, and polonius will compute the `region_live_at` relation based on that (right now, the `region_live_at` relation is an input to polonius).

We covered the overall plan in [this YouTube video](https://www.youtube.com/watch?v=i5KdU0ieb_A&list=PL85XCvVPmGQitE2CBzf-gERSqeXo59NQG&index=2&t=0s).

nikomatsakis:
Some immediate steps here:

- Extend [`AllFacts`](https://github.com/rust-lang/polonius/blob/4549e7123f8707123b9bd26621a3e8c0a0f15206/polonius-engine/src/facts.rs#L6) with a new type parameter `V: Atom` -- `V` stands for "local variable"
- Extend `AllFacts` with a field `var_used: Vec<(V, P)>` -- meaning "variable V is used at the point P" 
- Extend `AllFacts` with a field `var_defined: Vec<(V, P)>` -- meaning that the value in `V` is overwritten by the point P
- Get this to compile :)
- Extend the naive analysis with datafrog versions of the datalog rules below.

First, a variable is live on entry to P if it is used there:

```
var_live(V, P) :- var_used(V, P)
```

Next, a variable is live on entry to P if it is used in a successor Q and P does not reassign it:

```
var_live(V, P) :-
    var_live(V, Q),
    cfg_edge(P, Q),
    !var_defined(V, P).
```

At this point, we should be able to write some unit tests where we supply some base facts and a control-flow graph and then test `var_live` at those points.

nikomatsakis:
Once this is done, the next step will be to extend rustc to **produce** the `var_used` and `var_defined` facts.

nikomatsakis:
# Part 2 

Extending rustc to produce the `var_used` and `var_defined` facts.

First off, you will want to modify rustc to work with a local copy of polonius, so that you can edit the two in tandem. I think @albins you've already figured this part out so I'll just assume you have done so.

We're going to want to instantiate the polonius output with the "variable id" mapped to [the `mir::Local` type](https://doc.rust-lang.org/nightly/nightly-rustc/rustc/mir/struct.Local.html).

[...]

PR #105, opened 2019-03-27
Implement liveness computation

albins:
This is a work in progress on issue #104.

Sorry for the poorly named branch; I started it to fix a silly compiler warning and ended up using it for the implementation.


albins:
The current version seems to work, but I am not happy about generating an entire new `Relation` for walking the control-flow graph backwards.

I haven't yet written extensive tests, but I think it makes some sort of sense to go ahead with the extensions to rustc to produce the new facts so as to have some sort of ability to generate input data.

polonius-engine/src/output/naive.rs
```diff
@@ -179,17 +203,17 @@ pub(super) fn compute<Region: Atom, Loan: Atom, Point: Atom, Variable: Atom>(
...
+                    .or_insert_with(BTreeSet::new)
                     .insert(*borrow);
             }
 
             for (region, location) in &region_live_at_rel.elements {
                 result
                     .region_live_at
                     .entry(*location)
-                    .or_insert(vec![])
+                    .or_insert_with(|| vec![])
```
shepmaster: This should be `(Vec::new)` as below.

albins: Thanks! Fixed!

src/intern.rs
```diff
@@ -24,6 +24,10 @@ where
         &self.rev_strings[data]
     }
 
+    pub(crate) fn untern_vec(&self, data: &Vec<TargetType>) -> Vec<&str> {
```
shepmaster: https://stackoverflow.com/q/40006219/155423

albins: Thanks! I really appreciate these comments!

Changed data to be a slice of `TargetType`.

albins:
This is marked WIP again as I might as well roll the entire region logic into this PR as per #104.

PR #108, opened 2019-06-13
Add nice liveness graph output to the Polonius CLI wrapper

albins:
Uses Petgraph to reduce boring parts (completely linear sections with no changes in liveness of any variable) of the CFG; adds liveness edges with nice colours.

This only affects the Polonius CLI wrapper.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -413,7 +413,7 @@ dependencies = [
 
 [[package]]
 name = "polonius"
-version = "0.4.0"
+version = "0.5.0"
 dependencies = [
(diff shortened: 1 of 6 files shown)


shepmaster:
No pretty graphic to show off the results? boooooo 😉

albins:
@shepmaster: I forgot!
liveness

albins:
There is a problem with the labels of compacted nodes in the CFG: it shows them as first--last (when it's more than two), but it orders them by sorting the indices. Apparently, they are not guaranteed to translate to actually consecutive program points, but as far as I can tell there is almost nothing I can do about this.

albins:
Apparently, I snuck these changes in via another pull request for liveness changes. :)

PR #113, opened 2019-08-30
create a book

nikomatsakis:
For now I mostly stole content from the README.md. I'd like polonius to document its own inner workings and logic in here.


book/src/getting_started.md
```diff
@@ -0,0 +1,27 @@
...
+     Running `target/release/borrow-check 'inputs/clap-rs/app-parser-{{impl}}-add_defaults/'`
+--------------------------------------------------
+Directory: inputs/clap-rs/app-parser-{{impl}}-add_defaults/
+Time: 3.856s
+```
+
+You could also try `-a Naive` to get the naive rules (more readable,
+slower) -- these are the exact rules described in [the
+blogpost][post]. You can also use `-a LocationInsensitive` to use a
+location insensitive analysis (faster, but may yield spurious errors).
```
bjorn3: I wonder if trying `LocationInsensitive` and then `Naive` if the former fails would be faster overal.

book/book.toml
```diff
@@ -0,0 +1,5 @@
+[book]
+authors = ["Polonius authors"]
+multilingual = false
+src = "src"
+title = "Polonius"
```
bjorn3: Could the book be build on CI?

book/src/getting_started.md
nikomatsakis: that already exists, indeed. I think it's called `-a Hybrid`

book/book.toml
nikomatsakis: It could and it should be, but I was hoping to cajole somebody else into doing that :)

lqd: I'll take care of that

lqd:
(didn't we have bors here ?)

@bors r+

book/src/getting_started.md
lqd: Oh by the way, the [`Hybrid` variant](https://github.com/rust-lang/polonius/blob/master/polonius-engine/src/output/hybrid.rs) calls the optimized variant (`DatafrogOpt`) if it finds possible errors, not the `Naive` one.

This `Hybrid` variant is also the default one that `rustc` uses when calling into Polonius.

We have in-progress work which shares the results of the location insensitive pre-pass with the optimized variant, so that there are only a small number of possible errors to check (and that is a lot faster as well).

PR #114, opened 2019-08-30
Build the book on CI

lqd:
Try to build the mdbook on CI and autodeploy it from there

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,11 +19,7 @@ script:
 - cargo test --all
 - mdbook build ./book/
 
-deploy:
-  provider: pages
-  skip-cleanup: true
-  github-token: $GITHUB_TOKEN
-  local-dir: book/book
-  keep-history: false
-  on:
-    branch: master
+env:
+  RUSTINFRA_DEPLOY_DIR: book/book
+import:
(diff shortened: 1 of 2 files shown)


lqd:
According to [the doc](https://rust-lang-nursery.github.io/mdBook/continuous-integration.html) this _should be_ how it's done, what do you think @pietroalbini ? :)

pietroalbini:
We have a [different deployment code](https://github.com/rust-lang/infra-team/blob/master/guidelines/static-websites.md#configuring-travis-ci) for the rust-lang org. You need to replace the `deploy:` block with:

```yaml
env:
  RUSTINFRA_DEPLOY_DIR: book/book
import:
  - rust-lang/simpleinfra/travis-configs/static-websites.yml
```

lqd:
Ah thank you, I thought it was somewhat specific to `hugo` (as in the `compiler-team` repo), or at the very least unrelated to `mdbook` deployments since the set-up seemed different at first glance in the `rustc-guide` repo.

I'll change this ASAP.

pietroalbini:
That piece of infra is relatively new, so it's not surprising most of our repositories don't use it.

lqd:
Thanks a lot for the help @pietroalbini !

I guess it's time to ... try ? :)

PR #115, opened 2019-09-02
fix markup

tshepang:


lqd:
Thanks for the PR ! I had noticed this and it will be fixed once #114 lands

PR #122, opened 2019-09-09
an easy way to impl Default

tshepang:


polonius-engine/src/facts.rs
```diff
@@ -64,29 +64,6 @@ pub struct AllFacts<R: Atom, L: Atom, P: Atom, V: Atom, M: Atom> {
     pub path_accessed_at: Vec<(M, P)>,
 }
 
-impl<R: Atom, L: Atom, P: Atom, V: Atom, M: Atom> Default for AllFacts<R, L, P, V, M> {
```
nikomatsakis: We did not do this because `derive(Default)` adds the requirement that (e.g.) `R: Atom + Default`, which we don't want.

nikomatsakis:
I was going to close this PR, but actually @tshepang maybe you want to replace it with one that adds a comment for future folks to explain why the impl is there?

tshepang:
Why don't we want `Default` for this case (for my curiosity and for the upcoming PR)?

As an aside, I should have seen https://github.com/rust-lang/polonius/pull/49.

nikomatsakis:
@tshepang thanks! unfortunately this has a merge conflict... and, actually, with @csmoe's changes in #127, we actually probably **can** `derive(Default)` now

tshepang:
ok, will wait for integration into rustc to test the `derive(Default)`

nikomatsakis:
I think @tshepang we should be able to derive-default now, with perhaps some minor changes.

tshepang:
Is that more like it?

tshepang:
Ok, it still results in a failing rustc build (with [these changes](https://gist.github.com/lqd/03862f0f65a051fd60ef8074089b5c05))...

```
error[E0599]: no function or associated item named `default` found for type `polonius_engine::facts::AllFacts<borrow_check::nll::facts::RustcFacts>` in the current scope
   --> src/librustc_mir/borrow_check/nll/mod.rs:175:24                       
    |                                                                            
175 |         Some(AllFacts::default())                                                      
    |                        ^^^^^^^ function or associated item not found in `polonius_engine::facts::AllFacts<borrow_check::nll::facts::RustcFacts>`
    |                                                                  
    = note: the method `default` exists but the following trait bounds were not satisfied:
            `polonius_engine::facts::AllFacts<borrow_check::nll::facts::RustcFacts> : std::default::Default`
```

I will let this one rest.

PR #124, opened 2019-09-13
define the atoms

nikomatsakis:
I took a stab at some names.

- variables
- paths (nee MovePath)
- nodes (nee Points)
- loans
- origins (nee Regions)

Note that these all have a distinct first letter, should that be relevant, although I would also like to try and use written out words.

Thoughts? We can [haggle on Zulip](https://rust-lang.zulipchat.com/#narrow/stream/186049-t-compiler.2Fwg-polonius/topic/names.20of.20atoms) perhaps. =)

### DIFF
--- a/book/src/rules/atoms.md
+++ b/book/src/rules/atoms.md
@@ -1,1 +1,74 @@
-**To be written**
+Polonius defines the following **atoms**. To Polonius, these are
+opaque identifiers that identify particular things within the input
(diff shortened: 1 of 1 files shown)


albins:
...I now understand variables.

lqd:
if everyone likes `origins`, I've prepared https://github.com/rust-lang/polonius/pull/125 to at least start The Great Renaming™

PR #125, opened 2019-09-16
Clean up terminology: step 1 — rename `Region` to `Origin`

lqd:
A first step in cleaning up the Polonius terminology by:
- renaming `regions` into `origins` (except in datafrog closures and datalog rules comments, for now)
- whenever a single-letter generic type was used, using the full word instead
- assorted cleanups in comments or leftover random bits and bobs


nikomatsakis:
@lqd did you plan to add a few more commits here?

lqd:
@nikomatsakis I was thinking we could the renaming step by step, what do you think ? If others wanted to help, and so on.

nikomatsakis:
step by step is better!

PR #126, opened 2019-09-17
book: document input relations, naive rules, location-insensitive rules

lqd:
This PR starts describing the input relations "pre-liveness". More details about the liveness relations and atoms will be coming later. Also describes two of the three variants in datalog form.

Represents the state of things after #156 with the naming and rules from our [rules hackmd](https://hackmd.io/CGMNjt1hR_qYtsR9hgdGmw?view):
- added documentation about the input relations (with the new relations used in subset errors)
- added documentation about the `Naive` rules (_without_ the placeholder loans from the hackmd)
- added documentation about the `LocationInsensitive` rules
- some summary work about the rules and pipelines
- some links to blog posts and the polonius talk at Rust Belt Rust

r? @nikomatsakis


book/src/rules/relations.md
```diff
@@ -0,0 +1,137 @@
...
+let mut a = 0;
+let r = &mut a; // this creates the loan L0
+//      ^ let's call this 'a
+```
+
+there will be a `borrow_region` fact linking `L0` to `'a` at this node. This loan will flow along the CFG and the subset relationships between origins, and the computation will require that its terms are respected or it will generate an illegal access error.
+
+### 3. `universal_region`
+
+`universal_region(origin)`: this relation stores that the origin `origin` is a "universal region"/"free region"/"placeholder region" and not defined in the MIR body we're checking.
```
lqd: what terminology should be used here ?

book/src/rules/relations.md
```diff
@@ -0,0 +1,137 @@
...
+
+Those are the default universal origins (`'static`) and the ones defined on functions which are generic over a lifetime. For example, with
```
lqd: and here as well: should we call these universal origins, universal regions, free regions, free origins, etc ?

lqd:
(note that this uses the tentative CFG `nodes` terminology, which we probably won't adopt. When those  questions up there are more fleshed out, I'll switch this PR back to `points`)

book/src/rules/relations.md
```diff
@@ -0,0 +1,137 @@
...
+// `q` and `r` can be used here
+```
+
+the loan `L0` will be `killed` by the assignment, and this fact stored in the relation. When we compute which loans origins contain along the CFG, the `killed` nodes will stop this loan's propagation to the next CFG node.
+
+### 5. `outlives`
+
+`outlives(origin1, origin2, node)`: this relation stores that the origin `origin1` outlives origin `origin2` at the node `node`.
+
+This is the standard Rust syntax `'a: 'b` where the *lifetime* `'a` outlives the lifetime `'b`. From the node of view of origins as sets of loans, this is seen as a subset-relation: with all the loans in `'a` flowing into `'b`, `'a` contains a subset of the loans `'b` contains.
```
lqd: "from the node of view" is an overzealous s/point/node/g — which I'll fix when switching back to "points"

book/src/rules/relations.md
nikomatsakis: I regret the term universal region. Definitely not that.

I like the term *placeholder*, but in Rustc it is often used more narrowly, and *only* refers to the placeholders we create when processing "higher-ranked types" and "higher-ranked trait bounds".

nikomatsakis: I have been meaning to go and move the NLL logic to use "free regions" to refer to these names, with the premise that there is kind of a "binder" that binds all the names for regions that appear within the function body.

lqd: updated with placeholders and placeholders origins

book/src/rules/relations.md
lqd: typo fixed

lqd:
Also note that the syntax highlighter that comes with `mdbook` does not by default enable support for the prolog code blocks that I've used for the soufflé datalog rules: the md files will look a tiny bit better on github than they will in the deployed site (unless we also override the syntax highlighter to embed a version of the highlighter that enables prolog highlighting)

nikomatsakis (approved):

nikomatsakis:
I've skimmed this and it looks quite good. I'm inclined to merge it and give it more detailed reads later over time.

PR #127, opened 2019-09-18
Clean up terminology: step 2 — more renaming, and using actual words

lqd:
In which The Great Renaming ™ is basically done:

- continues the `region` to `origin` migration in the rest of `polonius-engine`
- renames `MovePath` to `Path`
- more consistent terminology around `loans` (preferred to `borrows` or `B`s)
- uses words instead of single-letters in comments, datalog rules, datafrog closures, etc basically everywhere, for all Atoms
- minor cleanups here and there

### DIFF
--- a/polonius-engine/src/facts.rs
+++ b/polonius-engine/src/facts.rs
@@ -4,64 +4,64 @@ use std::hash::Hash;
 /// The "facts" which are the basis of the NLL borrow analysis.
 #[derive(Clone, Debug)]
 pub struct AllFacts<T: FactTypes> {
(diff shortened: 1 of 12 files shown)


lqd:
However, since the names are now more descriptive, should I remove some of the "types" which were added to describe the single-letters variables, from e.g.
```
the origin `origin` may refer to data from loan `loan` starting at the point `point`
```
to something like
```
the `origin` may refer to data from `loan` starting at `point`
```
?

albins:
That sounds like a good way to make it more readable. :)

lqd:
rebased, squashed, and ready to go

Issue #111, opened 2019-08-30
replace "fact" type parameters with associated types


nikomatsakis:
At some point we should probably make a "dummy type" with associates types...

```rust
trait FactTypes {
    type Region: Atom;
    type Loan: Atom;
    type Path: Atom;
    type Variable: Atom;
}
```

etc

But not in this PR =)

_Originally posted by @nikomatsakis in https://github.com/rust-lang/polonius/pull/110_

PR #130, opened 2019-09-23
replace fact type parameters with FactTypes

csmoe:
Closes https://github.com/rust-lang/polonius/issues/111

### DIFF
--- a/polonius-engine/src/output/datafrog_opt.rs
+++ b/polonius-engine/src/output/datafrog_opt.rs
@@ -25,24 +25,24 @@ pub(super) fn compute<T: FactTypes>(
     let mut result = Output::new(dump_enabled);
 
     let var_maybe_initialized_on_exit = initialization::init_var_maybe_initialized_on_exit(
-        all_facts.child,
-        all_facts.path_belongs_to_var,
-        all_facts.initialized_at,
-        all_facts.moved_out_at,
-        all_facts.path_accessed_at,
-        &all_facts.cfg_edge,
+        all_facts.child.clone(),
+        all_facts.path_belongs_to_var.clone(),
+        all_facts.initialized_at.clone(),
+        all_facts.moved_out_at.clone(),
+        all_facts.path_accessed_at.clone(),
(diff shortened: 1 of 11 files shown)


polonius-engine/src/output/mod.rs
```diff
@@ -124,28 +117,21 @@ fn compare_errors<Loan: Atom, Point: Atom>(
...
-    Origin: Atom,
-    Loan: Atom,
-    Point: Atom,
-    Variable: Atom,
-    MovePath: Atom,
-{
+impl<T: FactTypes> Output<T> {
     pub fn compute(
-        all_facts: &AllFacts<Origin, Loan, Point, Variable, MovePath>,
+        all_facts: AllFacts<T>,
```
nikomatsakis: Why did this change from `&AllFacts` to `AllFacts`?

polonius-engine/src/output/datafrog_opt.rs
```diff
@@ -16,33 +16,33 @@ use crate::output::liveness;
...
 
     let var_maybe_initialized_on_exit = initialization::init_var_maybe_initialized_on_exit(
-        all_facts.child,
-        all_facts.path_belongs_to_var,
-        all_facts.initialized_at,
-        all_facts.moved_out_at,
-        all_facts.path_accessed_at,
-        &all_facts.cfg_edge,
+        all_facts.child.clone(),
+        all_facts.path_belongs_to_var.clone(),
```
nikomatsakis: Why do we now require all of these `clone` calls here?

albins: This might be related to the re-jiggering I am talking about: the initialisation function moves its inputs out of the struct as it's the only step using them (which is why I would have suggested splitting the struct). I'm not sure if the borrow checker isn't smart enough to see that they are only used in this function when this refactoring is applied? Either that, or something else. :)

csmoe:
clones cleaned up
r? @nikomatsakis

PR #133, opened 2019-10-11
Fix clap by updating the facts

lqd:
`clap` on master is missing some `.facts` files. 

PR #129 added warnings to ~nag~ remind us to refresh this dataset with up-to-date facts. Here they are.

Maybe adding a smoke test could be interesting, since it's an important dataset. The `LocationInsensitive` variant runs it in 700ms on my machine (1.5s on The Ole Laptop) so maybe it wouldn't be that bad a wait on `cargo test` (or if not, in a test subcrate: only `cargo test --all` would run it. And CI does that as well.)


shepmaster:
image

Wew lad

lqd:
That's a nice way of keeping my +/- ratio in the negatives — especially if I never add any tests ever again 👍

Issue #112, opened 2019-08-30
Break the analysis pipeline into components


albins:
Currently, the each Polonius "algorithm"'s `compute()` method performs initialisation analysis, then liveness calculations, and then the actual code. This means a lot of code duplication and in the case of the hybrid algorithm that all of this happens twice.

I suggest the following for a minimal modification to the current setup:
- the publicly visible `compute()` method still takes the gigantonormous `AllFacts` for now
- internally, these facts are split up into internal data structures consumed by the various `compute()` methods
- the calls initiating initialisation and liveness calculations happen in `output::compute()`

PR #134, opened 2019-10-12
Proposal: split `AllFacts` into contexts dedicated to each component of the pipeline

lqd:
This introduce contexts to store data between the pipeline's steps, to solve some of the problems described in #112.

These contexts are used to store:
- initialization data, used by liveness
- liveness data, used by the borrow checking variants
- some static input data, or partial results, which is shared between variants (and more could be shared)

This also makes more use of `Relation`s between initialization and liveness to avoid creating and cloning the same data (so a slight optimization here).

This separation should also drop the initialization input relations after `var_maybe_initialized_on_exit` is computed, and the liveness input relations after `region_live_at` is computed, instead of storing them all throughout the pipeline (although I haven't made sure of how much memory it "saves").

While this PR avoids some `Relation` cloning (but not all, as it'd require changes in concert with rustc, and which will be done in the future), it mostly avoids computing initialization and liveness twice as happens today, when using the `Hybrid` variant _and_ the `LocationInsensitive` pre-pass returns a potential error in the code. (These first 2 steps dominate runtime on bigger benchmarks when that happens)

The `Hybrid` variant also stores the potential errors found by the `LocationInsensitive` pre-pass, so that the following variant can heavily prune the data it traverses to find errors on only those specific "interesting" loans. (This is an optimization I've done elsewhere and has very good results on our small number of benchmarks, but can't upstream in a PR yet as this filtering could interfere with the ongoing work on illegal subset relations errors).

While sharing `Relation`s is easy when used as join inputs, it may also be interesting to share the static data inserted in the `Variable`s themselves, when appropriate (which is not that often right now) to avoid mapping and cloning some of this initial input data.

This also re-enables logging from within the `polonius` bin, which I had forgotten to push to the dependencies cleanup PR.


albins:
I think this looks great, but I'm not 100% comfy with the `mem::replace` thing. Specifically, I don't like the fact that those inputs are technically invalid after they have been consumed by the earlier stages, but that their access is no longer enforced by the compiler like it would have been if they were moved out. However, the only way to fix that (I guess) would have been to do the decomposition of `all_facts` on entry, which I guess has its own drawbacks?

lqd:
We can split it up more for sure because of the initial clone, while I was trying not to move away too much from AllFacts but there's no specific reason we couldn't do that. We also need to remove the clone and that will require changes in rustc as well, so I think we can use this to iterate on the kind of interface we'd need (after the clone), and then fix that upstream.

lqd:
(updated this to completely separate each step's data)

polonius-engine/src/output/mod.rs
```diff
@@ -81,54 +93,179 @@ pub struct Output<T: FactTypes> {
...
+        // 2) Liveness
+        let universal_regions = all_facts.universal_region.clone();
+
+        let liveness_ctx = LivenessContext {
+            var_used: all_facts.var_used.clone(),
+            var_defined: all_facts.var_defined.clone(),
+            var_drop_used: all_facts.var_drop_used.clone(),
+            var_uses_region: all_facts.var_uses_region.clone(),
+            var_drops_region: all_facts.var_drops_region.clone(),
+        };
```
albins: Why do you need these clones? I thought they were only used here?

lqd: For the same reason as the previous `mem::replace`: we can't move these out of `all_facts: &AllFacts<T>`

PR #136, opened 2019-10-15
stable toolchain works

tshepang:


tshepang:
Hm, is it not better to just remove the file?

lqd:
While there's no specific reason we can't use `stable`, the compiler uses the `beta` channel, and this is why we also have been using that. I believe it is because using that same version to avoid possible problems when updating rustc to newer polonius revisions. 

That being said, maybe I'm misremembering, and Niko may remember reasons to keep using beta, or the opposite, approve of switching to stable.

tshepang:
What compiler is meant in "the compiler uses the beta channel"? Rust compiler? I also do not understand "using that same version could avoid possible problems when updating rustc to newer polonius revisions".

lqd:
Ah exactly yes, the Rust compiler: this project is made specifically to be used used in rustc, in the [`librustc`](https://github.com/rust-lang/rust/tree/master/src/librustc) and [`librustc_mir`](https://github.com/rust-lang/rust/tree/master/src/librustc_mir) sub-crates. (Mainly the latter)

tshepang:
Ok, what do you mean by "the compiler uses the beta channel"?

bjorn3:
A nightly version of rust compiler is compiled using the latest beta version at that time. A beta version is compiled using the latest stable. A stable version is compiled using the previous stable version.

lqd:
rustc is built using bootstrapping: newer versions are built using an older version of itself, and the version it uses is the beta toolchain, the same as we use here today. There are more details in the rustc guide [here](https://rust-lang.github.io/rustc-guide/how-to-build-and-run.html#running-xpy-and-building-a-stage1-compiler) about the different stages.

tshepang:
interesting to know, thanks

PR #137, opened 2019-11-12
Illegal subset relation errors using placeholder loans

lqd:
Illegal subset relation errors were explained in [this blog post](http://smallcultfollowing.com/babysteps/blog/2019/01/17/polonius-and-region-errors/) and 2 solutions using the `subset` relation described there. This PR implements these errors using "placeholder loans" (a solution which was not described in the post). 

In this scheme, every "universal region" ("placeholder origin") also `requires` ("contains") a "placeholder loan", which flows into origins along the CFG like all the other loans. If at some point, one of these placeholder loans flows into a placeholder origin, we know there must be subset relation between the two. If this subset relation wasn't declared as a `known_subset`, then this is an error.

This is mostly implemented with this simple rule:

```prolog
subset_errors(origin1, origin2, point) :-
   requires(origin2, loan1, point),
   universal_region(origin2),
   placeholder_loan(origin1, loan1),
   !known_subset(origin1, origin2).
```

which can be read as prose as the earlier description: if the placeholder loan "loan1" from "origin1" flows into the placeholder origin "origin2", and it isn't declared that "origin1" is a subset of "origin2", then we need to emit a subset error.

Opening as draft for feedback, and:
- only the `Naive` variant has these errors for now, until we're a bit more sure this is a valid direction, and then I'll apply the technique to the `DatafrogOpt` variant
- I assumed the placeholder loans should be seeded into `requires` (AKA `contains`) at the root of the CFG, but getting this root uses an heuristic (the first edge in the `cfg_edge` relation) which probably holds for rustc and unit-tests, but not always for interned .facts files like we have in our input datasets, so maybe the root should be passed to the input with the other relations.
- computing these errors seems to prevent from using `LocationInsensitive` as a pre-pass filter in the `Hybrid` variant, or am I missing a fast way to make it work ?
- to add the placeholder loans data in the parser, I went with adding another additional header block to complement `universal_regions { ... }` instead of having a single block containing both (we can remove the former and rename the latter before landing this)
- I added simple dataset with 3 examples but it was partially manually constructed as I haven't added support for those facts to rustc yet, _and_ while there is a function where an expected subset error occurs, I'm not sure it's about the correct origins (the .facts being a bit distant from the original rust code, and rustfest-induced lack of sleep, make it hard to be sure right now)
- needs way more tests (possibly some from rustc could be imported)
- needs documentation in the book


albins:
I'm thinking we should really, really do something about our testing infrastructure in general. Testing correctness and understanding errors from correctness tests should not be this hard.

polonius-engine/src/output/mod.rs
```diff
@@ -76,6 +76,7 @@ impl ::std::str::FromStr for Algorithm {
 #[derive(Clone, Debug)]
 pub struct Output<T: FactTypes> {
     pub errors: FxHashMap<T::Point, Vec<T::Loan>>,
```
albins: Is this the point where we bite the bullet and rename errors, or do we do that further along?

albins:
>     * I assumed the placeholder loans should be seeded into `requires` (AKA `contains`) at the root of the CFG, but getting this root uses an heuristic (the first edge in the `cfg_edge` relation) which probably holds for rustc and unit-tests, but not always for interned .facts files like we have in our input datasets, so maybe the root should be passed to the input with the other relations.

One of the reasons I went with my current solution for move errors is specifically to avoid having to compute the root of the CFG, by the way.

polonius-engine/src/output/mod.rs
lqd: yeah good question

lqd:
> One of the reasons I went with my current solution for move errors is specifically to avoid having to compute the root of the CFG, by the way.

unless I'm mistaken about this PR's approach, that shouldn't too much of a problem: it seems well-known in rustc and in-memory facts in genera. Only in our input use-cases is this a "problem" (because of the interning) and there are many things we could do here (including just changing the `polonius`-bin to intern the CFG first).

polonius-engine/src/output/mod.rs
```diff
@@ -91,6 +91,7 @@ pub struct Output<T: FactTypes> {
     pub var_drop_live_at: FxHashMap<T::Point, Vec<T::Variable>>,
     pub path_maybe_initialized_at: FxHashMap<T::Point, Vec<T::Path>>,
     pub var_maybe_initialized_on_exit: FxHashMap<T::Point, Vec<T::Variable>>,
+    pub known_subset: FxHashMap<T::Origin, BTreeSet<T::Origin>>,
```
nikomatsakis: this doesn't seem useful in the new formulation, but maybe `known_contains`?

polonius-engine/src/output/mod.rs
nikomatsakis: I say do it

src/test.rs
```diff
@@ -501,3 +501,69 @@ fn var_drop_used_simple() {
...
 }
+
+/// This test ensures one of the two placeholder origins will flow into the
+/// other, but without declaring this subset as a `known_subset`, which is
+/// an illegal subset relation error.
+#[test]
+fn illegal_subset_error() {
+    let program = r"
+        universal_regions { 'a, 'b }
+        placeholder_loans { 'a: L1, 'b: L2 }
```
nikomatsakis: This feels tedious. I guess our new formulation would make this just 

```
placeholders { 'a: L1, 'b: L2 }
```

but I wonder if we can just write 

```
placeholders { 'a, 'b }
```

and use the `'a` names for the loans too?

src/test.rs
```diff
@@ -501,3 +501,69 @@ fn var_drop_used_simple() {
...
+
+    let result = Output::compute(&facts, Algorithm::Naive, true);
+
+    // ...so there should be an error here about the missing `'b: 'a` subset
+    assert_eq!(result.subset_errors.len(), 1);
+
+    let point = Point::from(1);
+    let subset_error = result.subset_errors.get(&point).unwrap();
+
+    let origin_a = Origin::from(0);
```
nikomatsakis: as an aside, a nice cleanup would be if we could produce "user-readable" errors and diff the output from those in these tests, no?

src/test.rs
lqd: yeah, that was the initial plan, this WIP state is to avoid most changes to `universal_regions` everywhere to focus on subset errors at first, and then clean this up

src/test.rs
lqd: yes I'd like this a lot, including reorganizing tests

lqd: I've made a couple attempts at this, and they're now more readable thanks to a `FactChecker` assertion helper. In https://github.com/rust-lang/polonius/pull/137/commits/0db817d4de435a28c21c863589a13d7cef8bf46a#diff-c5fc4beef0f03493c2e9a9c876073f0fR535 there's an example where the assertions use the regular atom values and not the interned ones.

src/test.rs
lqd: I've done this as we talked about: in the program syntax `placeholders { 'a, 'b }` replaces both `universal_regions { 'a, 'b }` and `placeholder_loans {'a: L1, 'b: L2}`, and is used to fill both the `universal_region` and `placeholder` input relations (until the redundant `universal_region` is removed everywhere, in rustc and the `.facts` files)

polonius-engine/src/output/mod.rs
lqd: done

lqd:
Since opening this PR, we've had some discussion in [this zulip thread](https://rust-lang.zulipchat.com/#narrow/stream/186049-t-compiler.2Fwg-polonius/topic/placeholder.20loans/near/181023543), and has resulted in these changes:
- `known_subset` was assumed to be "full" (ie, containing all the transitive results), but it's not the case in rustc, so we now compute a `known_contains` relation to ensure all placeholder loans are transitively contained by the placeholder origins in the known subset.
- instead of seeding the placeholder loans at the root of the CFG, which could stop propagating in subtle cases, we decided instead to make the placeholder origins contains these loans at all points of the CFG. The _effect_ would be the same, but "felt wrong" in previous cases for NLLs, so we thought it'd be better to be more conservative for now (and, possibly later, relax this after more thinking and testing).
- some more work was done on the parser and tests (and fact generation was implemented in a rustc branch)

Of note, since we decided to focus on correctness of the `Naive` variant at first before re-adapting our previous and ongoing optimization work, this PR [deactivates checking that the live loans are the same between variants](https://github.com/rust-lang/polonius/pull/137/commits/a6b609838ffb4f1cd73654b958147d7769b86337): the difference being that the placeholder loans are indeed live and the optimized variant doesn't have those.

As an additional note, I had quickly added the same subset error computation in the `DatafrogOpt` variant to not have to do this (or alternatively seeding `requires` with the loans everywhere would make this comparison pass), and it didn't produce the expected subset errors. I could have made a mistake doing so, or maybe the variant's TC avoidance is preventing some of these loans from flowing into `requires`, more investigation will be needed when we focus back on performance again.

With all this, I think this PR is ready for review.

polonius-engine/src/output/mod.rs
lqd: Let's do this at the same time we continue The Great Renaming of 2019 :)

polonius-engine/src/facts.rs
```diff
@@ -62,6 +62,14 @@ pub struct AllFacts<T: FactTypes> {
     /// `path_accessed_at(path, point)` when the `path` was accessed at point
     /// `point`. The same logic as for `initialized_at` and `moved_out_at` applies.
     pub path_accessed_at: Vec<(T::Path, T::Point)>,
+
+    /// `known_subset(origin1, origin2)` when the "placeholder" `origin1` is a known subset
+    /// of placeholder `origin2`.
```
nikomatsakis: This comment could be a bit better. I would say something like:

---

These reflect the `'a: 'b` relations that are either declared by the user on function declarations or which are inferred via implied bounds. For example, `fn foo<'a, 'b: 'a, 'c>(x: &'c &'a u32)` would have two entries, one for `'b: 'a` and one for `'a: 'c` (note that the transitive relation `'b: 'c` is not necessarily included explicitly, but rather inferred by polonius).

polonius-engine/src/output/mod.rs
```diff
@@ -213,22 +223,67 @@ impl<T: FactTypes> Output<T> {
...
+            Output::<T>::compute_known_contains(&known_subset, &all_facts.placeholder);
+
+        let placeholder_origin: Relation<_> = Relation::from_iter(
+            all_facts
+                .universal_region
+                .iter()
+                .map(|&origin| (origin, ())),
+        );
+
+        // TODO: flip the relation's arguments order in the input facts ?
```
nikomatsakis: yeah, we might as well make it match what's most useful

polonius-engine/src/output/naive.rs
```diff
@@ -68,6 +76,20 @@ pub(super) fn compute<T: FactTypes>(
                 .map(|&(origin, point)| ((origin, point), ())),
         );
 
+        // Placeholder loans are contained by their placeholder origin at all points of the CFG.
+        //
+        // contains(Origin, Loan, Node) :-
+        //   cfg_node(Node),
+        //   placeholder(Origin, Loan).
```
nikomatsakis: In the formal rules, we didn't have this, right? I'm fine with including it for now, though eventually (as we noted) we might want to try and refactor how we handle these "true everywhere" relations.

lqd: by "in the formal rules", do you mean the [canonical rules hackmd](https://hackmd.io/CGMNjt1hR_qYtsR9hgdGmw) ? If so, we did, in order to not seed the placeholder loans at the root of the CFG — but the comment indeed refers to the slightly older predicate name of:

```prolog
origin_contains_loan_on_entry(Origin, Loan, Node) :-
  cfg_node(Node),
  placeholder(Origin, Loan).
```
which is around the end of the [borrowck paragraph](https://hackmd.io/CGMNjt1hR_qYtsR9hgdGmw#Subset-and-contains-the-heart-of-the-borrow-check).

lqd: (and yeah it'd be great to refactor all the facts we materialize everywhere into a scalable and efficient solution)

nikomatsakis: OK, and yeah I meant the hackmd.

polonius-engine/src/output/mod.rs
lqd: Turns out we use both orders:
- computing `known_contains` uses the origin key to join with `known_subset`
- computing `subset_errors` uses the placeholder loan as a key: the join will return the "source placeholder" associated to a loan which we find flowing in a downstream origin

The comment was correct earlier in the PR's life but not anymore, I'll remove it.

PR #140, opened 2019-11-25
Bump versions for the subset errors computation

lqd:
This PR updates the versions so that we can update rustc to compute subset errors:
- the test `polonius-parser` crate now supports placeholder loans and known subsets
- the `polonius-engine` crate now supports the new input facts, new output errors, and computes subset errors where some outlives annotations might be required between free regions, for a function to be valid
- the `polonius` crate now supports the new facts for the `./inputs` datasets, and tests these new errors

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "polonius"
-version = "0.5.0"
+version = "0.6.0"
 authors = ["The Rust Project Developers", "Polonius Developers"]
 description = "Core definition for the Rust borrow checker"
 license = "Apache-2.0/MIT"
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "polonius"
-version = "0.5.0"
+version = "0.8.0"
```
shepmaster: Versions don't matter, but why 5 -> 8?

lqd: so IIUC, they were apparently documented as released in the releases.md file, but not actually released

lqd: I fixed the erroneous releases.md, and updated it to represent the current state of things, without missing an intermediate number, so this would be:
- `polonius-parser`: 0.4.0
- `polonius-engine`: 0.11.0
- `polonius`: 0.6.0

is that better ?

PR #144, opened 2020-03-06
Some move errors fixes and tests, and fact updates

lqd:
Some facts are missing on master, so this PR:
- makes missing fact files panic on load again
- re-adds the dataset used in tests to the fact collection script, and updates its facts
- adds a smoke test which tries to load the `clap` files, so that we know some files are missing
- regenerates the `clap` inputs to make it work again (however it has now become very slow, multiple minutes, and also shows false positives move errors)
- encode the 3 move errors tests in the `smoke-test` dataset into unit tests to have a bit of automated testing on move errors

Since this PR has a lot of fact files regenerated, it's best reviewed commit by commit with those fact regeneration commits ignored.

---
Update: I've added some commits which fix some move errors false positives on tracking move/init on paths and subpaths, and a test which would fail without it.

I've also located the source of the other false positives, in rustc fact generation, so after this PR lands and we publish a new release, I'll open the rustc PR, and then update the facts here again.

Also updates `polonius-engine` to 0.12.1 so we can publish that and work on the rustc PR.


polonius-engine/src/output/initialization.rs
```diff
@@ -49,7 +49,7 @@ fn compute_transitive_paths<T: FactTypes>(
     let path_begins_with_var = iteration.variable::<(T::Path, T::Variable)>("path_begins_with_var");
 
     // ancestor_path(Parent, Child) :- child_path(Child, Parent).
-    ancestor_path.insert(child_path.iter().collect());
+    ancestor_path.extend(child_path.iter().map(|&(child, parent)| (parent, child)));
```
albins: Ok that was a...stupid mistake.

PR #147, opened 2020-03-30
Update facts to remove the rest (🤞) of the move errors false positives

lqd:
PR #70546 has landed in rustc, fixing the fact generation issue causing some of the move errors false positives. Here's the PR updating our repository's facts with the updated fact gen.

As some of the false positives were encoded in the non-regression unit tests, we can more easily see, in the second commit, how the facts updates fix the issue. The rest of the changes are about moving the path accesses facts from the start point to the mid point. Not shown in the PR are the results on `clap`: the incorrect move errors are now gone.

(99% of this PR diff is from unrelated rustc changes which have also updated our `outlives` facts quite a bit; but I plan to to remove those specific facts soon to fix the rustc fact generation OOMs anyway.)


albins:
Looks good!

PR #150, opened 2020-06-26
Dump path_maybe_uninitialized_on_exit after compute_move_errors

utaal:
When called with --show-tuples and -v.

This may be useful to debug `move_error`s. If I missed the reason why this shouldn't be dumped, let me know!

### DIFF
--- a/polonius-engine/src/output/initialization.rs
+++ b/polonius-engine/src/output/initialization.rs
@@ -214,6 +214,14 @@ fn compute_move_errors<T: FactTypes>(
                 .or_default()
                 .push(path);
         }
+
+        for &(path, location) in path_maybe_uninitialized_on_exit.complete().iter() {
+            output
+                .path_maybe_uninitialized_on_exit
+                .entry(location)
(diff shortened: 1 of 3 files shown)


albins:
Thanks! Pretty sure it was a mistake on that part, but I’ll let the others chime in as well :)

lqd:
We probably haven't needed it just yet yeah :)

The verbose debug output can easily become overwhelming and hide the information you're looking for on the bigger datasets.

Thanks @utaal

utaal:
Makes sense. Thanks!

> The verbose debug output can easily become overwhelming and hide the information you're looking for on the bigger datasets.

What do you folks think of pre-pending the relation name to each fact on the debug output? This way when searching for an atom you can easily see all the facts (and relations) it appears in.

Something like this: https://gist.github.com/utaal/a88be7de8cbdd1e36c8a0a177489bdcf
A possible PR for this: https://github.com/rust-lang/polonius/compare/master...utaal:print-relation

PR #151, opened 2020-06-30
Print the relation name for each fact when dumping to stdout

utaal:
When writing output facts to stdout, it may be helpful to have the relation/variable name as the first column, so that one can search/grep the output for an atom and quickly find all relations/variables it appears in.

This is in part a possible answer to @lqd's comment on https://github.com/rust-lang/polonius/pull/150#issuecomment-650812589
> The verbose debug output can easily become overwhelming and hide the information you're looking for on the bigger datasets.

Stdout looks like this with this patch: https://gist.github.com/utaal/a88be7de8cbdd1e36c8a0a177489bdcf

If this isn't in line with your existing plans, let me know and I'll close.

### DIFF
--- a/src/dump.rs
+++ b/src/dump.rs
@@ -22,21 +22,22 @@ pub(crate) fn dump_output(
 ) -> io::Result<()> {
     macro_rules! dump_output_fields {
         ( $($field:ident),+ ) => {
-            $(dump_rows(
-                &mut writer_for(output_dir, stringify!($field))?,
-                intern,
-                &output.$field,
-            )?;)+
+            $({
+                let (name, mut write) = writer_for(output_dir, stringify!($field))?;
+                dump_rows(
+                    name,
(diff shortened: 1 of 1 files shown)


lqd:
Could be interesting. 

I myself was thinking about enabling/disabling specific relations via flags, but adding the relation names per row like here would work equally well.

Let's see what the others think

albins:
> Could be interesting.
> 
> I myself was thinking about enabling/disabling specific relations via flags, but adding the relation names per row like here would work equally well.
> 
> Let's see what the others think

Sounds like a good idea to me! We could always easily revert it if it turns out to be a bad idea anyway. :)

lqd:
CI is red presumably because of formatting, we can merge as soon as that's taken care of :)

lqd:
Thanks a lot @utaal

PR #155, opened 2020-12-20
Clean up terminology: step 3 — the rest of the owl

lqd:
Completes most of The Great Renaming™ to also align:

- input relations and .facts files (`borrow_region` to `loan_issued_at`, `killed` to `loan_killed_at`, `invalidates` to `loan_invalidated_at`, `outlives` to `subset_base`)
- some of the effects in the test parser mentioning the previous input relations
- the data computed by the variants ( `requires` relation to `origin_contains_loan_on_entry`, `borrow_live_at` to `loan_live_at`) and the API accessing this data
- some debug output relations (which used the very old `restricts` instead of `requires`, now use `origin_contains_loan_*`)
- (also removes an unused dataset)

with the naming plan in [our rules hackmd](https://hackmd.io/CGMNjt1hR_qYtsR9hgdGmw?view).

This PR will need require rustc's fact generation to be updated to use these new names as well.

For review purposes: since there's a lot of renaming, each commit can be reviewed individually. Note that the commits renaming the datasets' fact files are prefixed with "tests".


nikomatsakis (approved):

nikomatsakis:
this is amazing and you are a hero. do you want me to merge this *now*, @lqd, or do you have a particular ordering?

lqd:
@nikomatsakis this one was the one I planned we'd merge first :)

nikomatsakis:
DONE

PR #156, opened 2020-12-21
Compute subset errors everywhere

lqd:
This PR implements what we discussed last sprint:
- make the `Opt` variant compute subset errors, tracking the subsets of placeholder origins, because placeholder loans don't work in this situation
- make the `Naive` variant compute subset errors using the "easy subset-filtering" because we compute the full subset transitive closure here
- do nothing for now on the `LocationInsensitive` variant: placeholder loans are very natural in this variant, as we don't compute subsets at all
- by adding the subset errors to the `LocationInsensitive` variant, we can re-enable it as a pre-pass in the `Hybrid` variant: if no potential illegal access errors nor potential illegal subset relations errors are detected, there's no need to run the full analysis.

I've also started to number and document the rules, as preparation for putting them in the book (and extracting them as soufflé .dl files if need be). This can be seen in the `Naive` variant (commit 321c23e, before the computation was simplified), and if that's interesting I'll apply it to the other variants (though the `Opt` rules are already more documented than the 2 others).

As the last bit of renaming, it also aligns the relation name and input fact files of the `known_subset`s to what the finalized rules look like in [our hackmd](https://hackmd.io/CGMNjt1hR_qYtsR9hgdGmw?view#Known-contains) (now `known_placeholder_subset`s). The commits starting with "tests" are the ones renaming input files and can be glossed over/ignored.


lqd:
Working on documenting the `LocationInsensitive` variant in #126 made me realize I forgot to rename two relations in this PR here:
- `requires` was renamed to [`origin_contains_loan_on_entry`](https://github.com/rust-lang/polonius/blob/4a1e335c151c396473d9d79ab3eff56623a4a73c/polonius-engine/src/output/location_insensitive.rs#L39-L40) in the rules hackmd, but this name is specific to the `Naive` variant: this variant here does not track locations, and should be `origin_contains_loan`
- [`known_contains`](https://github.com/rust-lang/polonius/blob/4a1e335c151c396473d9d79ab3eff56623a4a73c/polonius-engine/src/output/mod.rs#L131) still uses its old name dual to [`known_subset`](https://github.com/rust-lang/polonius/blob/2cf8336f7ff9932270160a392ca5be3c804b7f41/polonius-engine/src/facts.rs#L69-L76) while the rules have it as `placeholder_known_to_contain`

I'll take care of both of these later, just to make my life easier until we have decided how to tackle #157, but I've used the correct names in the documentation PR.

polonius-engine/src/output/datafrog_opt.rs
```diff
@@ -378,6 +397,47 @@ pub(super) fn compute<T: FactTypes>(
...
+                    datafrog::ValueFilter::from(|&((origin1, _point), origin2), _| {
+                        origin1 != origin2
+                    }),
+                ),
+                |&((origin1, point), origin2), _| (origin1, origin2, point),
+            );
+
+            // subset_placeholder(Origin1, Origin3, Point) :-
+            //     subset_placeholder(Origin1, Origin2, Point),
+            //     subset(Origin2, Origin3, Point).
```
nikomatsakis, line 418: can we add a comment here that we maintain the invariant that `Origin1` is a placeholder region?

nikomatsakis (approved):

nikomatsakis:
r=me with nit

polonius-engine/src/output/datafrog_opt.rs
lqd: I've pushed a commit to address this

PR #163, opened 2021-05-24
typo fix

lengyijun:


lqd:
I thought I had fixed this one before, but it may have only been locally 😓 

Thank you!

PR #164, opened 2021-06-03
Switch from Travis to Github Actions

lqd:
Our CI has 👻-ed us, so let's try switching to GHA.

@Mark-Simulacrum does that look good to you ?

I've tested this a bit on my fork, and it seemed to work.

I'm a bit unsure about the book job: I don't have the admin rights to check if the actual tokens set up on this repo would keep working with the new simpleinfra job, or if new tokens or deploy keys would be needed.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,79 @@
+name: CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    
+
+env:
+  CARGO_TERM_COLOR: always
```
bjorn3, line 10: This is likely not necessary.

lqd: yes probably, it's what github has in its rust templates, so I left it in.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,79 @@
...
+          override: true
+          components: rustfmt
+
+      - name: Check formatting of all crates
+        run: cargo fmt --all -- --check
+
+  book:
+    name: Build and upload book
+    runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/master'
```
Mark-Simulacrum, line 35: FWIW moving this if to the "deploy book" step makes more sense - you likely do want to build things on PRs, just not deploy.

Mark-Simulacrum:
Not sure if GH CI isn't running on this PR for a particular reason or what's up with that - would probably be good to figure that out. Otherwise modulo one nit I left seems good to me, I would expect best way to test is to merge...

.github/workflows/ci.yml
lqd: That's true

lqd:
Maybe for the first GHA workflow has to be enabled via the UI... It's running now, so I'll be able to test the book parts.

lqd:
Seems ok up to actually deploying the book, using the tokens. So let's merge and test that. The currently deployed book is out of date anyways.

lqd:
Looking good

bjorn3:
Github actions is only activated for PRs once a workflow gets merged into the default branch.

PR #169, opened 2021-07-20
typo fix

lengyijun:


polonius-engine/src/facts.rs
```diff
@@ -60,7 +60,7 @@ pub struct AllFacts<T: FactTypes> {
     /// `path_is_var(path, var)` the root path `path` starting in variable `var`.
     pub path_is_var: Vec<(T::Path, T::Variable)>,
 
-    /// `path_assigned_at(path, point)` when the `path` was initialized at point
+    /// `path_assigned_at_base(path, point)` when the `path` was initialized at point
     /// `point`. This fact is only emitted for a prefix `path`, and not for the
     /// implicit initialization of all of `path`'s children. E.g. a statement like
     /// `x.y = 3` at `point` would give the fact `initialized_at(x.y, point)` (but
```
lqd, line 66: This is pre-existing but since you're modifying these comments, would you be able to also update the other similar old names references ? We can see `initialized_at` here and in other comments (which I assume should now be `path_assigned_at_base`, right @amandasystems ?),  and `moved_out_at` close below (probably `path_moved_at_base` here ?) and so on.

lengyijun: Ok. I will deal with it.

polonius-engine/src/facts.rs
```diff
@@ -60,19 +60,19 @@ pub struct AllFacts<T: FactTypes> {
...
     /// `point`. This fact is only emitted for a prefix `path`, and not for the
     /// implicit initialization of all of `path`'s children. E.g. a statement like
-    /// `x.y = 3` at `point` would give the fact `initialized_at(x.y, point)` (but
-    /// neither `initialized_at(x.y.z, point)` nor `initialized_at(x, point)`).
+    /// `x.y = 3` at `point` would give the fact `path_assigned_at_base(x.y, point)` (but
+    /// neither `path_assigned_at_base(x.y.z, point)` nor `path_assigned_at_base(x, point)`).
     pub path_assigned_at_base: Vec<(T::Path, T::Point)>,
 
     /// `path_moved_at_base(path, point)` when the `path` was moved at `point`. The
     /// same logic is applied as for `assigned_at` above.
```
lqd, line 71: here as well:
```suggestion
    /// same logic is applied as for `path_assigned_at_base` above.
```

lqd:
Thanks !

PR #173, opened 2021-07-30
Move to a fully hand-written parser to improve compile / iteration times

domenicquirl:
This is part of https://rust-lang.zulipchat.com/#narrow/stream/186049-t-compiler.2Fwg-polonius/topic/Polonius.20Hackathon.202021-07-30.

Preliminary results: 

```
---- ALL ----	   ---- OLD ----	---- NEW ----
build --all: 		   28.874		     8.38
build --all --release: 38.162			14.15
test: 				   42.74		    16.569
test --release:		   53.709			23.505

---- PARSER ONLY ----
build: 				   26.367		     1.427
build --release: 	   31.56			 2.84
test: 				   28.244			 2.1
test --release: 	   34.41			 2.945
```


lqd (commented):
Great job and great results! Thanks so much.

I left a couple of quick comments, but will try it out more later.

I would personally love to see a few more doc comments if you're up for it: it would really help understanding and maintaining this in the future (not that I expect the format to change any time soon).

For example, it would be great to have an extensibility example, either in the README or in the book, detailing the changes one would need to do to add, say, another effect emitting facts for a liveness relation. 

It _looks_ straightforward looking at the existing fact parsing in the PR, but since this is likely the most common operation we'll ever do on the parser, it seems like an interesting example to have.

polonius-parser/src/token.rs
```diff
@@ -0,0 +1,176 @@
...
+    Loan,
+    Variable,
+    Comment,
+    Whitespace,
+    Error,
+    Eof,
+}
+
+#[macro_export]
+macro_rules! T {
```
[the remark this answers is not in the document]
domenicquirl: `TokenKind` is just a `u16` and `Copy`, so there isn't really a need to intern it. The macro is more for convenience, and I also find it more readable since you have a lot of checks against kinds in `match`es and calls like `self.consume(T!['('])?;` or lists like
```rs
ParseError::UnexpectedToken {
    found,
    expected: vec![T![;], T![/]],
    position: self.position(),
}
```
A macro like this is used by the folks over at rust-analyzer, and has come in handy for me in my own projects as well a few times. I've added some explanation in the code that reflects what I'm explaining here.

domenicquirl:
@lqd I've made some changes addressing your comments. 

During the sprint, I was very much focused on getting this off the ground and compile times as far down as possible, largely flying by the existing tests. Now with more time I've done some minor refactorings to clean up the implementation, add documentation to a lot of places and give the `polonius_parser` crate its own README. The latter contains some general description and instructions, plus the example you asked about. 

I would have liked to put some doc tests on the actual parsing methods as well, but doc testing with internal items doesn't really work out that well. 

Let me know in case you have further questions on this.

polonius-parser/README.MD
```diff
@@ -0,0 +1,78 @@
+# Polonius Parser
```
bjorn3, line 1: Can you please use `.md` instead of `.MD` as extension?

domenicquirl: Sure, done.

lqd:
Thanks a ton!

polonius-parser/src/lexer.rs
```diff
@@ -0,0 +1,185 @@
...
+            }
+            kw if kw.starts_with("origin_live_on_entry".as_bytes()) => (
+                "origin_live_on_entry".len() as u32,
+                T![origin_live_on_entry],
+            ),
+            kw if kw.starts_with("var_dropped_at".as_bytes()) => {
+                ("var_dropped_at".len() as u32, T![var_dropped_at])
+            }
+            // effect keywords - use
+            kw if kw.starts_with("use".as_bytes()) => ("use".len() as u32, T![use]),
```
KvanTTT, line 126: Why prefix-like tree structure ([trie](https://en.wikipedia.org/wiki/Trie)) is not used here? It seems like it can provide additional performance because it considers all keywords simultaneously during comparison.

polonius-parser/README.md
```diff
@@ -0,0 +1,78 @@
...
+    goto B1, B2;
+}
+
+block B1 {
+    use('a), outlives('a: 'b), loan_issued_at('b, L1);
+}
+```
+
+## Usage
+The `polonius_parser` crate provides a single function `parse_input`, which takes a program description as its input string.
```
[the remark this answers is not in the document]
lqd: True, good catch. 

If you have a time to open a PR that would be great. Otherwise, I'll fix it soon.

polonius-parser/src/lexer.rs
lqd: Mostly because:
- the goal was to reduce compile times, and runtime is not especially important here: the tests are small, and not numerous yet. The slight gain in parsing could be interesting in the future, but not essential for this to land. If you're interested in benchmarking and improving this, by all means please do, we would love that.
- this whole PR was done during the latest friday-afternoon-sprint, and the lexer in particular was put together in a couple hours at most. Impressive work in such a short time.

PR #174, opened 2021-08-01
Prepare release 0.13.0

lqd:
This PR prepares a `polonius-engine` release:
- removes the warnings for the 2 currently unused context sharing fields for the `Hybrid` variant: the potential errors that the `LocationInsensitive` variant computes can (and will, in the future) be used by the `Naive`/`DatafrogOpt` variants to limit their own computation to the loans/subsets that actually matter.
- updates the version to 0.13.0, which includes the results of the more recent round of renaming, and changes to compute subset errors in all variants: this will allow the `Hybrid` variant to be the default once again.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -331,7 +331,7 @@ dependencies = [
 
 [[package]]
 name = "polonius-engine"
-version = "0.12.1"
+version = "0.13.0"
 dependencies = [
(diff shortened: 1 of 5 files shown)


lqd:
Testing this branch under rustc looks good to me so I'll merge and do the release.

PR #177, opened 2021-08-12
Filter reflexive subset relations eagerly

ecstatic-morse:
Depends on rust-lang/datafrog#37.

Filter reflexive `subset` relations *before* they are added to the `recent` tuples list.


polonius-engine/src/output/datafrog_opt.rs
```diff
@@ -299,11 +280,14 @@ pub(super) fn compute<T: FactTypes>(
...
-            //   dying_can_reach_live(origin2, origin3, point1, point2).
-            subset_o1p.from_join(
+            //   dying_can_reach_live(origin2, origin3, point1, point2),
+            //   origin1 != origin3.
+            subset_o1p.from_join_filtered(
                 &live_to_dying_regions_o2pq,
                 &dying_can_reach_live,
-                |&(_origin2, _point1, point2), &origin1, &origin3| ((origin1, point2), origin3),
+                |&(_origin2, _point1, point2), &origin1, &origin3| {
+                    (origin1 != origin3).then(|| ((origin1, point2), origin3))
```
lqd, line 289: Personally not a big fan of `bool::then`, but I don't mind if you and the others like it.

polonius-engine/src/output/datafrog_opt.rs
```diff
@@ -418,11 +396,14 @@ pub(super) fn compute<T: FactTypes>(
...
-            //     subset(Origin2, Origin3, Point).
-            subset_placeholder.from_join(
+            //     subset(Origin2, Origin3, Point),
+            //     Origin1 != Origin3.
+            subset_placeholder.from_join_filtered(
                 &subset_placeholder_o2p,
                 &subset_o1p,
-                |&(_origin2, point), &origin1, &origin3| (origin1, origin3, point),
+                |&(_origin2, point), &origin1, &origin3| {
+                    (origin1 != origin3).then(|| (origin1,  origin3, point))
```
lqd, line 405:
```suggestion
                    (origin1 != origin3).then(|| (origin1, origin3, point))
```
startup idea: rustfmt-as-a-service

lqd:
This is how I envisioned we would do this in the end, so this looks good to me once we publish a datafrog release including https://github.com/rust-lang/datafrog/pull/37.

Do you remember if such predicates are accepted by Soufflé btw ? So that we can potentially extract datalog rules from our comments/book and easily run them with it (which is one of the reason they are (intended to be) written with upper-cased variables for example, and not the rust-like naming that they used to have). 

Doesn't matter much if it doesn't support it, maybe we just add the symmetry-removing predicate as a comment rather than part of the rule, or not.

ecstatic-morse:
Souffle supports inequality as a predicate, [along with many others](https://souffle-lang.github.io/constraints), so that shouldn't be an issue. I like that it makes the rules correspond directly to the `datafrog` code.

polonius-engine/src/output/datafrog_opt.rs
ecstatic-morse: I quite like it. If you and/or Niko find it too obtuse, we can change it of course.

lyxell:
I don't know the context here, but FYI, Soufflé has no requirements on the case of variables, you can even use the same identifier for the type and the attribute: i.e. `.decl errors(loan: loan, point: point)` is valid Soufflé.

ecstatic-morse:
Force pushed with `rustfmt` changes applied. There were some missing trailing commas as well.

PR #178, opened 2021-08-12
Don't re-index relations just to add an empty tuple

ecstatic-morse:
Depends on rust-lang/datafrog#36.

Removes some indexes, which should help memory usage by a small amount and makes things easier to read.

This seems to make `clap-rs` a bit slower for some reason, which needs to be investigated before this can be merged.

### DIFF
--- a/polonius-engine/src/output/datafrog_opt.rs
+++ b/polonius-engine/src/output/datafrog_opt.rs
@@ -30,13 +30,10 @@ pub(super) fn compute<T: FactTypes>(
         let loan_killed_at = &ctx.loan_killed_at;
         let known_placeholder_subset = &ctx.known_placeholder_subset;
         let placeholder_origin = &ctx.placeholder_origin;
+        let loan_invalidated_at = &ctx.loan_invalidated_at;
 
         // Create a new iteration context, ...
         let mut iteration = Iteration::new();
-
-        // `loan_invalidated_at` facts, stored ready for joins
-        let loan_invalidated_at =
-            iteration.variable::<((T::Loan, T::Point), ())>("loan_invalidated_at");
 
         // `loan_issued_at` input but organized for join
         let loan_issued_at_op =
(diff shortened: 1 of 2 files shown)


lqd:
The reason why rules 7 and 8 can't be merged yet is another interesting datafrog limitation that could be of interest to you :)

lqd:
Apart from the above point (which we could keep as a comment for the time being, and fix some day in the future): needless to say this looks good to me once we have a datafrog release with the required PR.

ecstatic-morse:
> The reason why rules 7 and 8 can't be merged yet is another interesting datafrog limitation that could be of interest to you :) It's about leapjoins' well-formedness and there are details in the comment about why we used to need both relation and variable versions of the liveness facts.

Hmm, looking at this again, it seems like it's more of an indexing issue? `origin_live_on_entry` and `loan_invalidated_at` are both relations, but they don't have a shared key or something?

lqd:
Merging rules 7 and 8 would be something like

```prolog
errors(Loan, Point) :-
    loan_invalidated_at(Loan, Point),
    origin_contains_loan_on_entry(Origin, Loan, Point),
    origin_live_on_entry(Origin, Point).
```

If this were implemented as filter leapers on `origin_contains_loan_on_entry` it would be a leapjoin that is not well-formed: IIRC what Frank said (and I don't think it's documented anywhere ?), leapjoins' WF-ness currently requires at least one extend leaper. 

This criterion manifests as runtime errors rather than compile-time errors: IIRC related to the MIN/MAX sentinel indices in the leapjoins, like an assertion that at least one leaper returned a valid index. I think the errors can be triggered on some of the in-repo input tests, with the two rules merged; at least they used to when we tried this 3y ago. And there are some discussions with Frank about this on Zulip where he mentioned a way to fix it in datafrog (it was some kind of passthrough leaper just for that purpose). I can find these discussions if you want.

ecstatic-morse:
No need to dig through the chat logs. That's pretty clear to me. I've always been a bit confused by the difference between "extend" and "filter" (though more the "anti" variants). "extend" is for when you have terms in the leaper that are not bound by the `Variable`, correct?

lqd:
Exactly

PR #183, opened 2022-01-16
Rewrite `polonius-engine`

ecstatic-morse:
Requires rust-lang/datafrog#50.

Each section ("naive", "initialization", etc.) is now expressed as a
`Computation`, with defined inputs and outputs. Computations are
organized into `Pipeline`s, which execute each computation in turn,
storing the results in a `Db` where they are accessible to later
computations in the pipeline.

`Computation`s can also dump intermediate results using the
(creatively named) `Dump` struct. A series of `Dumper`s can be specified
for a pipeline, each which will be given access to a type-erased
version of the dumped data.


ecstatic-morse:
I'm not sure what the path is to upstream this. It's too big to review on its own, obviously. I might have to work in my fork for a while.

polonius-engine/src/pipeline.rs
```diff
@@ -0,0 +1,221 @@
...
+    info!("Running computation `{}`...", name);
+
+    let input = <C::Input<'_>>::load_from_db(db);
+    dump.unit_start(name);
+    let start_time = Instant::now();
+    let output = computation.compute(input, dump);
+    let end_time = Instant::now();
+    dump.unit_end(name);
+
+    output.store_to_db(db, dump);
```
ecstatic-morse, line 209: Here's the part that requires GATs/associated type bounds. Moving the lifetime parameter for `Input` to `Computation` itself and changing the signature to the following gives an error:

```rust
fn compute_<'db, C, T>(computation: &C, db: &'db mut Db<T>, dump: &mut Dump<'_>)
where
    C: Computation<'db, T, Input: LoadFrom<'db, T>, Output: StoreTo<T>>,
    T: FactTypes,
```

```
error[E0502]: cannot borrow `*db` as mutable because it is also borrowed as immutable
   --> polonius-engine/src/pipeline.rs:209:5
    |
193 | fn compute_<'db, C, T>(computation: &C, db: &'db mut Db<T>, dump: &mut Dump<'_>)
    |             --- lifetime `'db` defined here
...
202 |     let input = <C::Input>::load_from_db(db);
    |                 ----------------------------
    |                 |                        |
    |                 |                        immutable borrow occurs here
    |                 argument requires that `*db` is borrowed for `'db`
...
209 |     output.store_to_db(db, dump);
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ mutable borrow occurs here
```

This is expected since data from `db` flows from `input` into `computation` and then `output`. It's perhaps more surprising that this works with a GAT on `Computation::Input`?

ecstatic-morse: My understanding is that this works with GATs because neither the implementer of `Computation` nor `Computation::Output` can name the lifetime used in `Computation::Input<'db>`. I'm not very confident in this assessment, however.

dullbananas:
@ecstatic-morse are you still working on this?

PR #179, opened 2021-08-19
Add an optional Soufflé backend

ecstatic-morse:
Allows `polonius` to execute Datalog programs directly using [Soufflé](https://souffle-lang.github.io/). 

In the long term, that means:
   - We can test that the results of our datafrog computations are correct, even intermediate ones like `subset`.
   - We can prototype new ideas more quickly. No need to do stratification or re-indexing by hand before seeing results.

This PR binds to [Soufflé's C++ interface](https://souffle-lang.github.io/interface) using [`cxx`](https://github.com/dtolnay/cxx). As a result, we can initialize relations and extract results entirely in memory; no need to write intermediate results to disk. If I were to start from scratch, I might choose generate Soufflé binaries instead of libraries and serialize everything on `tmpfs`. It took a lot of work to get `cxx`, `rustc`, and the generated C++ code to play nicely.

For now, this exposes a single `--souffle` flag that will use the Soufflé backend instead of the existing `datafrog` one. This is just a proof-of-concept. There's a lot of work to do before it can be merged:

- [ ] Put this behind a feature flag.
- [ ] Make sure the rules are correct.
    - I just used copies that I had on an old branch.
- [ ] Make sure that the results of the Soufflé backend are correct.
   - There were discrepancies in the few tests that I ran locally. This may be due to fact serialization, outdated rulesets, or something else entirely.
- [ ] Integrate into the existing fact dumping infrastructure.
- [ ] Update the "rules" section of the book now that we generate markdown files from the datalog files directly (see `literate.py`).
- [ ] Other assorted FIXMEs (see the source)

PR #184, opened 2022-01-20
Rewrite the doc generator in Rust

frank-king:
This PR rewrites the automatic doc generator in #179 in Rust.

It is a draft version yet, since the script is directly translated from the original Python script, and the generated doc are not checked yet.


frank-king:
Close it since no time to work on it.

PR #193, opened 2023-07-21
Make Polonius report any found errors by default

ivanbakel:
It's a bit confusing that running Polonius on some Rust code that you know has a borrow error doesn't seem to actually do anything - it gives the impression that no analysis is performed by default.

This changes the behaviour to print error tuples by default if any are found, for a better experience. The old behaviour can be re-enabled by using `--no-show-tuples` to suppress tuple output even if errors are present.


hkBst:
Seems like an improvement.

PR #198, opened 2024-06-19
Enrich gitignore file to exclude editor files

FullyNonlinear:
When using CodeLLDB on vscode to debug, vscodes files are currently tracked. This PR add editor files to gitignore so that editor files are no longer tracked by git.


hkBst:
Ideally I think you should ignore editor files in your personal global git config.

PR #200, opened 2025-06-06
deploy book via official actions

marcoieni:
Related to https://github.com/rust-lang/simpleinfra/issues/597

The infra team would like to get rid of the action `rust-lang/simpleinfra/github-actions/static-websites` so we are replacing it with the official GitHub actions.

- [x] After receiving an approval, I need to change Pages settings before merging this

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,13 +4,13 @@ on:
   push:
     branches: [ master ]
   pull_request:
-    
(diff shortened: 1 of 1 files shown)


lqd (approved):

marcoieni:
oops, there's an error. fixing in https://github.com/rust-lang/polonius/pull/201

